PR #3, opened 2017-09-07
Set up multi-stage docker build.

mindprince:


Dockerfile
```diff
@@ -12,8 +12,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM gcr.io/google_containers/cos-nvidia-driver-install@sha256:456f272257d4557518afcaa7b0a1bba22d6861b20061ed5a1bd3f38aa7f8fb1a
+FROM golang:1.8.3-stretch as builder
```
vishh: What does "stretch" mean?

Dockerfile
```diff
@@ -12,8 +12,12 @@
...
+WORKDIR /go/src/github.com/GoogleCloudPlatform/container-engine-accelerators
```
vishh: Where does the source code gets pulled? Are you expecting to use a volume as part of docker build?
Will it just work on cloud builder?
My understanding was this is why we had to use the cloud builders image workflow.

Dockerfile
mindprince: Debian Stretch (9)

Dockerfile
mindprince: This is just setting the [`WORKDIR`](https://docs.docker.com/engine/reference/builder/#workdir), I could have set it to `/anything` and it would work.

PR #5, opened 2017-09-14
Update to use the version which doesn't do git pull

mindprince:


mindprince:
Once the container builder builds the new image, I will update the daemonset to use that sha.

vishh:
lgtm

PR #6, opened 2017-09-15
Use init container for installing the driver. Minor code cleanup

mindprince:
Instead of installing them from inside the device plugin.
This makes it easier to control these two aspects separately.
We also switch to using the prepackaged installer image.
The resulting device-plugin image would be much smaller.
Need to update the device-plugin-gpu image sha once it is generated by container builder.
Also, minor code cleanup.


mindprince:
Tested this by building the image in my GCR repo, updating the daemonset with that URL, triggering the device-plugin e2e test locally. It passed.

PR #12, opened 2017-10-16
Updating device plugin support to v1alpha2 API

vishh:
Once https://github.com/kubernetes/kubernetes/pull/53031 merges I'll update vendor deps and then this PR should be good to go.


vishh:
This PR is ready to go. I think it can be merged even before  kubernetes/kubernetes#53031 merges

device-plugin-daemonset.yaml
```diff
@@ -74,7 +74,7 @@ spec:
         - name: sysrq
           mountPath: /sysrq
       containers:
-      - image: "gcr.io/google-containers/device-plugin-gpu@sha256:dbd0fc7087f6831200840bbd5b39699981a9b4ce98ffb3eea8174f20f1d6ab2b"
+      - image: "gcr.io/google-containers/device-plugin-gpu@sha256:84ca5cfefef891c9354a9a1cecda0dc89f5682e710afbf0689b8f541fb754dd2"
```
mindprince: Let's update this in a separate PR.

This tag for this sha currently points to a commit (`12940a9e11bf234763dc5950a0c8016dbc4ef139`) which doesn't exist. I would prefer if I can look at the sha, find the tag for it in GCR and see which commit in this repo it belongs to.

vishh: Ack

mindprince:
Can you update this, so that the tests start passing?

vishh:
Removed `_vendor` directory

PR #15, opened 2017-10-21
Supports device hot plugin.

jiayingz:
Instead of assuming nvidia gpu devices are already properly set up
when device plugin starts, wait till nvidia gpu devices appear.
This change is needed to run nvidia gpu device plugin as a system
addon that is automatically deployed on nodes with gpu devices,
while driver installation is still triggered by users and happen
at later time.


mindprince:
A minor improvement maybe to make the time configurable.

PR #18, opened 2017-10-25
Logging, comments and other improvements.

mindprince:
- Use glog instead of fmt. Need to run with `-logtostderr` to see log in stderr.
- Better log messages.
- Minor comment fixes.
- Remove redundant `return` and `os.Exit` after `glog.Fatal`.
- Some IDE suggested fixes and more consistency in naming.


jiayingz:
Overall lgtm. Are you going to change device-plugin-daemonset.yaml file to add -logtostderr argument to device plugin command?

mindprince:
Yeah, I will do that when we update it with the new image.

PR #20, opened 2017-10-26
Stop the server before fatally exiting.

mindprince:


cmd/nvidia_gpu/nvidia_gpu.go
```diff
@@ -144,7 +145,6 @@ func (ngm *nvidiaGPUManager) ListAndWatch(emtpy *pluginapi.Empty, stream plugina
 			if dev.Health != state {
 				changed = true
 				dev.Health = state
-				ngm.devices[id] = dev
```
jiayingz: Without this, would the actual value in ngm.devices[id] be updated with the new state?

mindprince: Wow. TIL that range loop creates copies.

PR #24, opened 2017-11-07
a small bug fix in nvidia_gpu_test.go

WIZARD-CXY:
Move ```defer conn.Close()``` after ```as.Nil(err)``` in case a grpc.Dial error


WIZARD-CXY:
@jiayingz PTAL

PR #25, opened 2017-11-14
No need to run device-plugin manually anymore.

mindprince:
It's an addon in GCP now. The manifest for that is here:
https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/device-plugins/nvidia-gpu/daemonset.yaml

Instead of that a pause container is run, so that the installer can remain an init container.
Also, added node affinity to only run on nodes that have a label with key 'cloud.google.com/gke-accelerator'.

---

There are three files now:
1) device-plugin-daemonset.yaml
2) daemonset.yaml
3) daemonset-preloaded.yaml

(1) already existed but its name no longer makes sense. (2) is copy of (1).
(3) is copy of (2) except that the installer image is assumed to be present on
the node. (1) would be deleted once tests in kubernetes/kubernetes are updated
to point to (2) or (3).


daemonset-preloaded.yaml
```diff
@@ -0,0 +1,83 @@
...
+          path: /proc/sysrq-trigger
+      initContainers:
+      - image: "cos-nvidia-installer:fixed"
+        imagePullPolicy: Never
+        command: ["/bin/sh", "-c"]
+        args: ["usr/bin/nvidia-installer.sh"]
+        name: nvidia-driver-installer
+        resources:
+          requests:
+            cpu: 0.15
```
jiayingz: Nit: should we increase this based on what we measured in PR https://github.com/kubernetes/kubernetes/pull/53541 that installer sometimes can take up to 2 cores?

mindprince: I am not sure. That would in effect reduce the capacity of all GPU nodes by 2 cores.

jiayingz: I also have similar concern. What I am worried about is installer may take much longer time when the node is short of cpu. Perhaps we can discuss this later outside this PR.

/lgtm

PR #26, opened 2017-11-27
Update version of Jupyterhub to 0.8.1

foxish:


foxish:
@vishh, maybe the overall image itself should also be simplified for now?

vishh:
LGTM

Yeah, the image needs to be trimmed and simplified. The base image is itself 1+ GB in size. I haven't gotten around to that yet.

PR #27, opened 2017-11-27
Adding jupyterhub configuration

foxish:
PTAL @vishh 
cc @kow3ns

This should directly work with the GPU image after https://github.com/GoogleCloudPlatform/container-engine-accelerators/pull/26

### DIFF
--- a/jupyterhub/manifests/config.yaml
+++ b/jupyterhub/manifests/config.yaml
@@ -1,0 +1,109 @@
+# Copyright 2017 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: v1
+data:
+  jupyterhub_config.py: |
+    import json
+    import os
+    from kubespawner.spawner import KubeSpawner
+    from oauthenticator.github import GitHubOAuthenticator
+
+    class KubeFormSpawner(KubeSpawner):
+      def _options_form_default(self):
+        return """
+        <label for="image">Image</label>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;
+        <input name="image" placeholder="repo/image:tag"></input>
+        <br/><br/>
+
+        <label for="cpu_guarantee">CPU</label>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;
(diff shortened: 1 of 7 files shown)


example/jupyterhub.yaml
```diff
@@ -0,0 +1,62 @@
...
+  - port: 80
+    name: http
+    targetPort: 8000
+  selector:
+    app: tf-hub
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: tf-hub-0
```
vishh: Why is this service named as it is?

example/jupyterhub.yaml
```diff
@@ -0,0 +1,62 @@
...
+  replicas: 1
+  template:
+    metadata:
+      labels:
+        app: tf-hub
+    spec:
+      containers:
+      - name: tf-hub
+        image: gcr.io/google-containers/k8s-jupyterhub:0.1
+        imagePullPolicy: Always
```
vishh: Why is it always?

example/jupyterhub.yaml
foxish: This is because the hub pods expect to connect back on tf-hub-0:8081. We could work around this and supply the fully qualified statefulset name, but that would need more configuration to be passed through. Chose this in the interest of simplicity.

vishh: Why is it `-0`?

example/jupyterhub.yaml
vishh: ping @foxish

jupyterhub/manifests/config.yaml
```diff
@@ -0,0 +1,109 @@
...
+    # Don't try to cleanup servers on exit - since in general for k8s, we want
+    # the hub to be able to restart without losing user containers
+    c.JupyterHub.cleanup_servers = False
+    ###################################################
+
+    ###################################################
+    ### Spawner Options
+    ###################################################
+    c.JupyterHub.spawner_class = KubeFormSpawner
+    c.KubeSpawner.singleuser_image_spec = 'foxish/tf-nb-1'
```
vishh: can this be updated to `gcr.io/kubeflow/tensorflow-notebook`?

jupyterhub/manifests/config.yaml
```diff
@@ -0,0 +1,109 @@
...
+    c.KubeSpawner.cmd = 'start-singleuser.sh'
+    c.KubeSpawner.args = ['--allow-root']
+    # First pulls can be really slow, so let's give it a big timeout
```
vishh: Can we setup a job with anti-affinity to pull images on all nodes? May be as a next step?

mindprince:
Also please squash the commits.

example/jupyterhub.yaml
foxish: This is already fixed in the new version.

jupyterhub/manifests/config.yaml
foxish: Will open an issue. Can do that post-kubecon?

jupyterhub/manifests/config.yaml
foxish: Done

foxish:
Also updated image to `gcr.io/kubeflow/jupyterhub:1.0`

PR #28, opened 2017-11-29
kubernetes model server

kow3ns:


.idea/vcs.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
vishh: Did you mean to check this in?

k8s-model-server/README.md
```diff
@@ -0,0 +1,170 @@
...
+
+## Create Google Cloud Platform (GCP) Project 
+This tutorial uses [GCP](https://cloud.google.com/).
+[This documentation](https://cloud.google.com/resource-manager/docs/creating-managing-projects) demonstrates how to 
+create and manage a new Project on GCP. There is a [free tier](https://cloud.google.com/free/) available that you can 
+use to try it out.
+
+## Build a Model Server Docker Image
+The [docker file](docker/Dockerfile) builds a docker image containing default the TensorFlow model server tuned for 
+amd64 architecture. You can modify this file to support the architecture of your choice.
```
vishh: As discussed today, can we use a GPU compatible image as the default?

k8s-model-server/README.md
```diff
@@ -0,0 +1,170 @@
...
+
+```
+ENV MS_USER=model-server
+```
+
+This is the user that the model-server process will run as. You can modify this user to change the name. This user 
+will be used in the Kubernetes Deployment's PodSecurityContext so that the container is not run with root privileges
+(This is considered to be a best practice).
+
+### Build the Image
```
vishh: Can we have a `summary` section in the beginning where we recommend using an image that we have already compiled? Compiling manually is most likely for different platform optimizations or for private registry reasons.

k8s-model-server/README.md
```diff
@@ -0,0 +1,170 @@
...
+
+Use the [gsutil mb](https://cloud.google.com/storage/docs/gsutil/commands/mb) command to create a bucket. Note that 
+the bucket name must be globally unique.
+```commandline
+gsutil mb gs://<ucket-name>
+``` 
+
+
+## Upload a Model
+You can train an inception model using the instructions from the
```
vishh: Mention that inception is an example, but the workflow is adaptable to any TF model. It currently reads as though the guide is only for inception.

k8s-model-server/README.md
```diff
@@ -0,0 +1,170 @@
...
+gs://<bucket-name>/inception/1/saved_model.pb
+
+gs://<bucket-name>/inception/1/variables/:
+gs://<bucket-name>/inception/1/variables/variables.data-00000-of-00001
+gs://<bucket-name>/inception/1/variables/variables.index
+```
+
+## Create a Kubernetes Cluster
+
+Use [gcloud container clusters create](https://cloud.google.com/sdk/gcloud/reference/container/clusters/create) to turn
```
vishh: Why not link to GKE documentation? I'd also include a GPU guide here. Serving TF using CPUs is probably better suited for Lambda or App engine as long as the model fits within the resource constraints. I feel GPU or on-prem is where k8s shines.

k8s-model-server/README.md
```diff
@@ -0,0 +1,170 @@
...
+        image: gcr.io/<your-project>/model-server:1.0
+```
+
+You need to update the Deployment's `spec.template.spec.containers[0].command.args` to point the GCS bucket you created  
+above.
+
+```yaml
+ args:
+        - /usr/bin/tensorflow_model_server
+          --port=9000 --model_name=inception --model_base_path=gs://<your-bucket>/inception/
```
vishh: What happens if no models exist? Would the server crashloop or watch for updates?

k8s-model-server/docker/Dockerfile
```diff
@@ -0,0 +1,43 @@
+FROM ubuntu:16.04
+
+MAINTAINER Kenneth Owens <kowens@google.com>
```
vishh: This API is deprecated. Suggest replacing as a label.

k8s-model-server/manifests/model-server.yaml
```diff
@@ -0,0 +1,64 @@
...
+      containers:
+      - name: model-server
+        image: gcr.io/kubeflow/model-server:1.0
+        imagePullPolicy: Always
+        resources:
+          requests:
+            memory: "1Gi"
+            cpu: "1"
+          limits:
+            memory: "4Gi"
```
vishh: This will not run on minikube most likely. Can we run as BestEffort to begin with and recommend placing resource limits to users?

k8s-model-server/manifests/model-server.yaml
```diff
@@ -0,0 +1,64 @@
...
+  labels:
+    run: model-service
+  name: model-service
+spec:
+  ports:
+  - port: 9000
+    targetPort: 9000
+  selector:
+    app: model-server
+  type: LoadBalancer
```
vishh: Wouldn't it better to have an ingress instead? L4 is simpler, but realistically one would use an ingress right as serving deployments come and go.

.idea/vcs.xml
kow3ns: Nope

k8s-model-server/README.md
kow3ns: Sure but where are we hosting that image. We can update later once the image gets pushed.

k8s-model-server/README.md
kow3ns: +1

k8s-model-server/README.md
kow3ns: Slowly crash loops.

k8s-model-server/manifests/model-server.yaml
kow3ns: I don't think ingress is fully compatible with GRPC. We can consider updating to an ingress later after testing it.

k8s-model-server/README.md
vishh: gcr.io/kubeflow?

k8s-model-server/README.md
vishh: So by default, if we install TF Serving, then it will keep crashlooping until a model is made available?

vishh:
Merging, let's iterate as we go.

mindprince:
```
remote: warning: File k8s-model-server/inception.tar.gz is 96.32 MB; this is larger than GitHub's recommended maximum file size of 50.00 MB
```

@kow3ns @vishh Was this file meant to be checked in?

PR #29, opened 2017-11-30
Adding bazel to notebook image

vishh:
Moving images to kubeflow GCP project

cc @foxish


vishh:
Tested that this works. Merging. Happy to address any comments later.

PR #30, opened 2017-11-30
Make PD accessible and install bazel as root

vishh:
cc @foxish This includes bazel for real and inception training works with this change now.


example/tensorflow-notebook.yaml
```diff
@@ -44,12 +44,11 @@ spec:
       - key: nvidia.com/gpu
         effect: NoSchedule
       securityContext:
-        supplementalGroups:
-        - 0
+        runAsUser: 0
```
foxish: It needs to run as privileged?

vishh: Good catch, I did not mean to check that in.

vishh: Fixed PTAL

PR #31, opened 2017-11-30
Kubeflow documentation

foxish:
Did a broad pass at the documentation. Lots of TODOs. PTAL at the overall structure as well.

cc/ @vishh @jlewi @kow3ns @aronchick

### DIFF
--- a/kubeflow/README.md
+++ b/kubeflow/README.md
@@ -1,6 +1,6 @@
 # Kubeflow Anywhere
 
-This document detals the steps needed to run the kubeflow project in different environments, such as Minikube (local laptop). Google Kubernetes Engine, etc. 
+This document details the steps needed to run the kubeflow project in different environments, such as Minikube (local laptop). Google Kubernetes Engine, etc. 
 
 ## Quick Start
 
@@ -10,9 +10,9 @@ In order to quickly setup all components of the stack, run:
 kubectl apply -f manifests/ -R
 ```
 
-The above command sets up JupyterHub, an API for training using Tensorflow, and a set of deployment files for serivng. 
+The above command sets up JupyterHub, an API for training using Tensorflow, and a set of deployment files for serving. 
 Used together, these serve as configuration that can help a user go from training to serving using Tensorflow with minimal
-effort in a portable fashion between different environments. You can refer Instructions for using each of these components below. 
+effort in a portable fashion between different environments. You can refer to instructions for using each of these components below. 
 
 ## Setup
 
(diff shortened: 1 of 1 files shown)


kubeflow/README.md
```diff
@@ -0,0 +1,117 @@
+# Kubeflow Anywhere
+
+This document detals the steps needed to run the kubeflow project in different environments, such as Minikube (local laptop). Google Kubernetes Engine, etc.
```
jlewi: details

kubeflow/README.md
```diff
@@ -0,0 +1,117 @@
+# Kubeflow Anywhere
```
jlewi: Should there be a brief intro/overview explaining what Kubeflow is?

kubeflow/README.md
```diff
@@ -0,0 +1,117 @@
...
+
+In order to quickly setup all components of the stack, run:
+
+```commandline
+kubectl apply -f manifests/ -R
+```
+
+The above command sets up JupyterHub, an API for training using Tensorflow, and a set of deployment files for serivng. 
+Used together, these serve as configuration that can help a user go from training to serving using Tensorflow with minimal
+effort in a portable fashion between different environments. You can refer Instructions for using each of these components below.
```
jlewi: "refer to instructions"

aronchick:
I'm happy to take a swing today.

kubeflow/README.md
```diff
@@ -0,0 +1,117 @@
...
+
+```commandline
+kubectl apply -f manifests/ -R
+```
+
+The above command sets up JupyterHub, an API for training using Tensorflow, and a set of deployment files for serivng. 
+Used together, these serve as configuration that can help a user go from training to serving using Tensorflow with minimal
+effort in a portable fashion between different environments. You can refer Instructions for using each of these components below. 
+
+## Minikube
```
jlewi: Do we need to have top level sections for minikube and GKE which then have the same subections (e.g. single node, distributed training).

The message we want to send is that the commands on minkube are the same as GKE.

So ideally we had have a single set of instructions for the various steps that should work regardless of the deployment.

For training this should work; i.e. to launch a training job the command should be

kubectl create -f tf_job.yaml



/cc @aronchick

foxish: I considered that, and switched between layouts a bit. Open to suggestions.

foxish: Done, changed layout to have less duplication.

vishh:
This is a good start. @aronchick can you do some wordsmithing to get this to a reasonable shape. This should be the Readme in the google/kubeflow repo.

vishh:
@foxish is this still relevant?

foxish:
This can be closed. It's been moved to https://github.com/google/kubeflow

PR #32, opened 2017-11-30
[k8s-1.8] Mount the whole directory where nvidia libraries and binaries are present.

mindprince:
The old installer had the libraries inside HostPathPrefix/lib and we were
mounting them as ContainerPathPrefix/lib64. The new installer has libraries
inside HostPathPrefix/lib64. To work-around this discrepancy, mount the whole
HostPathPrefix as ContainerPathPrefix.

The one potentially breaking affect of this change is when using the old
installer, the libraries inside the container will now be available as
/usr/local/nvidia/lib instead of /usr/local/nvidia/lib64. However, I tested
with the test cuda image and things worked fine.

Also, make hostPathPrefix and containerPathPrefix command-line flags, so that
they can be used with images other the COS.


mindprince:
(Corresponding PR for master https://github.com/GoogleCloudPlatform/container-engine-accelerators/pull/33)

PR #35, opened 2017-12-01
Use nightly tf package and include tf benchmarks

vishh:


vishh:
cc @foxish

example/tensorflow-notebook.yaml
```diff
@@ -47,10 +47,10 @@ spec:
         fsGroup: 0
       containers:
       - name: tf-notebook
-        image: gcr.io/kubeflow/tensorflow-notebook
+        image: gcr.io/kubeflow/tensorflow-notebook-cpu:v1
         resources:
-          limits:
-            nvidia.com/gpu: 1 # Number of GPUs
+#          limits:
```
foxish: Should this just be removed entirely?

vishh: It's there as an example.

foxish:
LGTM

PR #39, opened 2017-12-16
Add nvidia-driver-installer for ubuntu that works the same way as the one for COS.

mindprince:
Tested it on the upcoming ubuntu image for GKE 1.9

### DIFF
--- a/nvidia-driver-installer/ubuntu/Dockerfile
+++ b/nvidia-driver-installer/ubuntu/Dockerfile
@@ -1,0 +1,23 @@
+# Copyright 2017 Google Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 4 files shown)


mindprince:
Also, here's the diff between the new cos entrypoint.sh and ubuntu's entrypoint.sh: https://gist.github.com/anonymous/fc89e481d9529cfe5744d4b3a19df64c

vishh:
This looks good to me, but I'd prefer having someone from Canonical review this. Perhaps @yguo0905 can help with this.

mindprince:
Also, @cmluciano was interested in Ubuntu support IIRC. Maybe he can also test it.

(I have tested it on an yet unreleased ubuntu image that will be used with GKE 1.9)

yguo0905:
If we need a formal review, we should create an internal ticket to explain what we do and ask someone from Canonical to take a look.

cmluciano:
Excellent! I will try this out.

mindprince:
@cmluciano Note that this is just the driver installer. After this, you would have to run https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/device-plugins/nvidia-gpu/daemonset.yaml for the device plugin.

PR #40, opened 2017-12-16
Use the new installer for COS that updates ldconfig on host.

mindprince:
Only merge this once https://github.com/kubernetes/kubernetes/pull/57255 is merged.

### DIFF
--- a/daemonset.yaml
+++ b/daemonset.yaml
@@ -11,14 +11,9 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-#
-# Yaml file to install a DaemonSet that contains a device plugin for
-# managing nvidia gpu devices. Expected to be deployed in Kubernetes
-# clusters with nvidia gpu devices. On each node, the DaemonSet detects
-# whether there are nvidia gpu devices, and if so, sets up the devices
-# in ready state and report them to Kubelet through the device plugin
-# interface. The device plugin simply sleeps if there is no nvidia
-# gpu device on the node.
+
+# The Dockerfile and other source for this daemonset are in
+# https://github.com/GoogleCloudPlatform/cos-gpu-installer
 
 apiVersion: extensions/v1beta1
 kind: DaemonSet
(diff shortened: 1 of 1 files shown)


daemonset.yaml
```diff
@@ -44,39 +39,36 @@ spec:
...
-      - name: sysrq
+      - name: root-mount
         hostPath:
-          path: /proc/sysrq-trigger
+          path: /
       initContainers:
-      - image: "gcr.io/google-containers/cos-nvidia-driver-install@sha256:de2d9098df423b064bb2579c4fb114022b2a6b7cc39c677d7ac720c9b75d9eb5"
-        command: ["/bin/sh", "-c"]
-        args: ["usr/bin/nvidia-installer.sh"]
+      - image: gcr.io/cos-cloud/cos-gpu-installer:v20171215
```
vishh: Can we use a SHA please?

daemonset.yaml
```diff
@@ -44,39 +39,36 @@ spec:
...
           privileged: true
         env:
-          - name: BASE_DIR
-            value: "/rootfs/nvidia"
-          - name: DEVICE_PLUGIN_ENABLED
-            value: "true"
+          - name: NVIDIA_DRIVER_VERSION
+            value: "384.81"
+          - name: NVIDIA_INSTALL_DIR_HOST
+            value: /home/kubernetes/bin/nvidia
```
vishh: Din't we decide to change this to `/var/lib/nvidia` or something like that? @Amey-D

mindprince: We can set this to `/var/lib/nvidia` but then I have to update the device-plugin as well.

daemonset.yaml
```diff
@@ -44,39 +39,36 @@ spec:
...
+          - name: NVIDIA_INSTALL_DIR_CONTAINER
```
vishh: Please add comments for what each of these env vars are meant for

daemonset.yaml
mindprince: Also, I think that requires some more hoops: https://github.com/GoogleCloudPlatform/cos-gpu-installer/blob/master/install-test-gpu.cfg#L41

daemonset.yaml
mindprince: Done.

daemonset.yaml
mindprince: Because there are some other variables as well, I think it makes more sense to document them at the script level where all of them are present. cc - @Amey-D

PR #42, opened 2017-12-19
Match gRPC socket path in Kubelet

ScorpioCPH:
This PR use `MountPath` which matched Kubelet

@vishh @mindprince PTAL. Thanks!

### DIFF
--- a/cmd/nvidia_gpu/nvidia_gpu.go
+++ b/cmd/nvidia_gpu/nvidia_gpu.go
@@ -43,7 +43,7 @@ const (
 	nvidiaDeviceRE       = `^nvidia[0-9]*$`
 
 	// Device plugin settings.
-	pluginMountPath      = "/device-plugin"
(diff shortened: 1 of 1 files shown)


mindprince:
This path is the path inside the container, it points to the correct path on the host.

https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/device-plugins/nvidia-gpu/daemonset.yaml#L32-L34
https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/device-plugins/nvidia-gpu/daemonset.yaml#L52-L53

ScorpioCPH:
@mindprince Hi, thanks for your reply, but how to deploy this by binary?
I test this on local without container, this path is not work.

mindprince:
If you want to run this directly without a container, maybe symlink `/device-plugins` to `/var/lib/kubelet/device-plugins`?

ScorpioCPH:
ok, sgtm.

PR #51, opened 2018-01-28
Add support to Debian 9

bhack:
I've not actually a Debian host with Nvidia GPU to test it.
Have you a K8 Debian 8 GPU node to test it?

### DIFF
--- a/cmd/nvidia_gpu/README.md
+++ b/cmd/nvidia_gpu/README.md
@@ -18,4 +18,9 @@ Examples of how driver installation needs to be done can be found at:
   - Installer code: https://github.com/GoogleCloudPlatform/container-engine-accelerators/blob/master/nvidia-driver-installer/ubuntu/entrypoint.sh
   - Installer daemonset: https://github.com/GoogleCloudPlatform/container-engine-accelerators/blob/master/nvidia-driver-installer/ubuntu/daemonset.yaml
 
+- For [Debian](https://cloud.google.com/kubernetes-engine/docs/concepts/node-images#debian) (experimental):
+  - Installer code: https://github.com/GoogleCloudPlatform/container-engine-accelerators/blob/master/nvidia-driver-installer/debian/entrypoint.sh
+  - Installer daemonset: https://github.com/GoogleCloudPlatform/container-engine-accelerators/blob/master/nvidia-driver-installer/debian/daemonset.yaml
+
+
 In short, this device plugins expects that all the nvidia libraries needed by the containers are present under a single directory on the host. You can specify the directory on the host containing nvidia libraries using `-host-path`. You can specify the location to mount that directory in all the containers using `-container-path`. For example, let's say on the host all nvidia libraries are present under `/var/lib/nvidia/lib64` and you want to make these libraries available to containers under `/usr/local/nvidia/lib64`, then you would use `-host-path=/var/lib/nvidia/lib64` and `-container-path=/usr/local/nvidia/lib64`.
--- a/nvidia-driver-installer/debian/daemonset.yaml
+++ b/nvidia-driver-installer/debian/daemonset.yaml
@@ -1,0 +1,73 @@
+# Copyright 2017 Google Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: extensions/v1beta1
+kind: DaemonSet
+metadata:
+  name: nvidia-driver-installer
+  namespace: kube-system
+spec:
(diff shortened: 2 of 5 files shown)


cmd/nvidia_gpu/README.md
```diff
@@ -18,4 +18,9 @@ Examples of how driver installation needs to be done can be found at:
   - Installer code: https://github.com/GoogleCloudPlatform/container-engine-accelerators/blob/master/nvidia-driver-installer/ubuntu/entrypoint.sh
   - Installer daemonset: https://github.com/GoogleCloudPlatform/container-engine-accelerators/blob/master/nvidia-driver-installer/ubuntu/daemonset.yaml
 
+- For [Debian](https://cloud.google.com/kubernetes-engine/docs/concepts/node-images#debian) (experimental):
```
mindprince: This link is 404.

nvidia-driver-installer/debian/daemonset.yaml
```diff
@@ -0,0 +1,73 @@
...
+        hostPath:
+          path: /dev
+      - name: nvidia-install-dir-host
+        hostPath:
+          path: /home/kubernetes/bin/nvidia
+      - name: root-mount
+        hostPath:
+          path: /
+      initContainers:
+      - image: gcr.io/google-containers/debian-nvidia-driver-installer@sha256:7ffaf40fcf6bcc5bc87501b6be295a47ce74e1f7aac914a9f3e6c6fb8dd780a4
```
mindprince: This is wrong. Point it to `latest` for now.

mindprince:
Also, can you please squash your commits?

cmd/nvidia_gpu/README.md
bhack: I cannot find a Google reference about Debian node image docs.

nvidia-driver-installer/debian/daemonset.yaml
bhack: So do you want a cut before `@sha`?

bhack:
Done.

mindprince:
Thinking more about this, since everything is the same as the ubuntu script, would it make more sense to just add notes in the README that the ubuntu script/image works for Debian as well? It would also make sure that the ubuntu and debian scripts don't go out of sync.

bhack:
It is not the same cause the base docker image is changing. I.e. check [this line](https://github.com/GoogleCloudPlatform/container-engine-accelerators/pull/51/files#diff-573da5fd7bbb10c8ede2e8be0162ab84R43?)

bhack:
What I  mean is that for the kernel headers the host node and the installing container os versions needs "to match". Basically for a Debian 9 node we need a Debian 9 installing container.

bhack:
An alternative is to introduce in ubuntu build args for the `Dockerfile`, arg variable for `Makefile`, and a variable for the `daemonset.yaml`

mindprince:
> It is not the same cause the base docker image is changing. I.e. check [this line](https://github.com/GoogleCloudPlatform/container-engine-accelerators/pull/51/files#diff-573da5fd7bbb10c8ede2e8be0162ab84R43)

> What I mean is that for the kernel headers the host node and the installing container os versions needs "to match". Basically for a Debian 9 node we need a Debian 9 installing container.

I have yet to test it, but I don't think the base docker image being Ubuntu would be a problem.
 `uname -r` returns the host kernel version, so all we need from the container is that the apt repositories it is looking at contains that kernel version. If debian is using one of the kernel versions that are part of the ubuntu apt repositories, everything would work as expected. If not, all we need to do is to update the [ubuntu Dockerfile](https://github.com/GoogleCloudPlatform/container-engine-accelerators/blob/f56351f7b2fc48e75b3c52e5f4ae101c9aa02f0c/nvidia-driver-installer/ubuntu/Dockerfile#L17) to add the apt repository which contains the debian kernel.

> An alternative is to introduce in ubuntu build args for the Dockerfile, arg variable for Makefile, and a variable for the daemonset.yaml

This maybe preferable if using the same docker image doesn't work out.

bhack:
Generally.the kernels mismatch but let me know when you have a Debian node with Gpu to test it.

vishh:
What's the maintenance plan? I don't think we at Google have bandwidth to maintain additional distros at the moment.

bhack:
@vishh How do you want to solve https://github.com/kubernetes/kubernetes/issues/54011? There is still Debian in the title.

vishh:
@bhack If you can setup a test infra for this feature, we can consider taking it. Otherwise, we may have to wait for Nvidia to release their driver installer container images

bhack:
I don't own test infra. What do you mean? Also this [new Nvidia announce](https://news.developer.nvidia.com/kubernetes-on-nvidia-gpus-release-candidate-now-available/) is only marketing or there is something new?

vishh:
How do you plan on testing this PR continually against different Debian
releases?
 @3XX0 might be able to share more about Nvidia's driver installation
containers. He had demo'ed it previously in a k8s WG meeting.

bhack:
Are we talking of https://github.com/NVIDIA/nvidia-container-runtime/?

bhack:
I see in the distro matrix at https://nvidia.github.io/nvidia-container-runtime/ that Debian is covered. So what we need to do?

flx42:
@bhack look at this document:
https://docs.google.com/presentation/d/1NY4X2K6BMaByfnF9rMEcNq6hS3NtmOKGTfihZ44zfrw/edit?usp=sharing

PR #54, opened 2018-02-09
Add nvidia-driver-installer for CoreOS Container Linux

lsjostro:
This PR adds nvidia driver installer for CoreOS Container Linux.

### DIFF
--- a/nvidia-driver-installer/containerlinux/Dockerfile
+++ b/nvidia-driver-installer/containerlinux/Dockerfile
@@ -1,0 +1,23 @@
+# Copyright 2017 Google Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+FROM ubuntu:17.10
+
+RUN apt-get update && \
+    apt-get install -y kmod gcc make curl xz-utils bc libelf-dev libssl-dev && \
+    rm -rf /var/lib/apt/lists/*
+
+COPY entrypoint.sh /entrypoint.sh
+
+CMD /entrypoint.sh
--- a/nvidia-driver-installer/containerlinux/Makefile
+++ b/nvidia-driver-installer/containerlinux/Makefile
@@ -1,0 +1,25 @@
+# Copyright 2017 Google Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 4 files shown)


lsjostro:
I signed it!

discordianfish:
@lsjostro I'm trying this out right now. This only addresses the driver parts though, right? It's doesn't seem to install the libnvidia-container nor nvidia-container-runtime, correct?

lsjostro:
@discordianfish correct.  We are using googles own [nvidia k8s device plugin](https://raw.githubusercontent.com/kubernetes/kubernetes/release-1.9/cluster/addons/device-plugins/nvidia-gpu/daemonset.yaml) instead of the official nvidia plugin, which doesn't require a custom docker runtime. Works really well for us.

Currently we have coreos nvidia installer Dockerfile hosted [here](https://github.com/shelmangroup/coreos-gpu-installer) and docker image [here](https://hub.docker.com/r/shelmangroup/coreos-nvidia-driver-installer/)

discordianfish:
@lsjostro Ah thanks for that hint! So this installer container and the k8s device plugin should be enough?

I thought it's needed since my pods still fail to schedule (Insufficient nvidia.com/gpu). I'm using the same k8s device plugin manifest. Guess I've been down the wrong route.

When dropping the `nvidia.com/gpu: 1` limit, the [test pod](https://gist.github.com/discordianfish/9ac14ec94f13bd3c7cc42c94482b04a6) gets scheduled but fails with:

```
libdc1394 error: Failed to initialize libdc1394
..
cudaRuntimeGetVersion() failed with error #35
```

The k8s plugin is running fine though:
```
I0213 10:52:04.891238    1412 nvidia_gpu.go:258] device-plugin started
I0213 10:53:35.096858    1412 nvidia_gpu.go:82] Found Nvidia GPU "nvidia0"
I0213 10:53:35.096935    1412 nvidia_gpu.go:209] starting device-plugin server at: /device-plugin/nvidiaGPU-1518519215.sock
I0213 10:53:35.097257    1412 nvidia_gpu.go:231] device-plugin server started serving
I0213 10:53:35.098269    1412 nvidia_gpu.go:240] device-plugin registered with the kubelet
I0213 10:53:35.098918    1412 nvidia_gpu.go:143] device-plugin: ListAndWatch start
I0213 10:53:35.098948    1412 nvidia_gpu.go:159] ListAndWatch: send devices &ListAndWatchResponse{Devices:[&Device{ID:nvidia0,Health:Healthy,}],}
```

Well, guess I have to dig deeper. Looks like the installer worked fine. Thanks a lot for that!

discordianfish:
I think with the approach of bind-mounting the shared libs from the host, once has to run ldconfig to update the ld cache before starting an application. The official digits container isn't doing that.

lsjostro:
make sure you add `-host-path=/opt/nvidia` to the k8s device plugin cmdline.

discordianfish:
Yep, figured that out. Maybe the default should be changed, /home/kubernetes is an odd choice.

vishh:
I'd recommend having a single daemonset perform driver installation and run the device plugin and that way sharing driver artifacts are controlled via a single config file. 

On the other hand, we at Google do not have bandwidth to setup CI and maintain installations for additional OSes. I would like to not merge this until we identify a maintenance plan.

lsjostro:
@vishh thanks for the feedback! I totally understand that! We’ll host it here in the meanwhile https://github.com/shelmangroup/coreos-gpu-installer

amoghkashyap:
@lsjostro Thanks for the patch. Nvidia-drivers are installed on the V100. However, I am having difficulty in getting the runtime set to NVIDIA. Because of this, I am not able to run nvidia-docker or use the nvidia-smi inside the Kubernetes pod which needs GPU acceleration. Request you to help me out.

CoreOS version : 2079.3.0

Also, I guess the docker tag has to be used as "master" in the daemonset.yml since "latest" is not being pulled.

lsjostro:
@amoghkashyap sure! mind coping the issue to https://github.com/shelmangroup/coreos-gpu-installer ?

PR #55, opened 2018-02-17
Add support for v1beta1 device plugin API.

jiayingz:
Other than unit tests, also ran GPUDevicePlugin e2e test with and without PR https://github.com/kubernetes/kubernetes/pull/58282 to make sure the gke gpu device plugin works with both v1alpha kubelet and v1beta1 kubelet.


Makefile
```diff
@@ -37,8 +37,10 @@ presubmit: vet
 TAG?=$(shell git rev-parse HEAD)
 REGISTRY?=gcr.io/google-containers
 IMAGE=nvidia-gpu-device-plugin
+REPO="github.com/GoogleCloudPlatform/container-engine-accelerators"
 
 build:
+	go install ${REPO}/pkg/gpu/nvidia
```
vishh: You don't need to install

Dockerfile
```diff
@@ -15,6 +15,7 @@
 FROM golang:1.9-alpine as builder
 WORKDIR /go/src/github.com/GoogleCloudPlatform/container-engine-accelerators
 COPY . .
+RUN go install github.com/GoogleCloudPlatform/container-engine-accelerators/pkg/gpu/nvidia
```
vishh: Why is this necessary?

vishh:
Fix the `use_alpha` logic as discussed offline and then this is good to be merged. Feel free to self-merge.

pkg/gpu/nvidia/manager.go
```diff
@@ -0,0 +1,212 @@
...
+	nvidiaCtlDevice = "/dev/nvidiactl"
+	nvidiaUVMDevice = "/dev/nvidia-uvm"
+	// Optional device.
+	nvidiaUVMToolsDevice = "/dev/nvidia-uvm-tools"
+	devDirectory         = "/dev"
+	nvidiaDeviceRE       = `^nvidia[0-9]*$`
+)
+
+var (
+	hostPathPrefix      = flag.String("host-path", "/home/kubernetes/bin/nvidia", "Path on the host that contains nvidia libraries. This will be mounted inside the container as '-container-path'")
```
vishh: nit: Flags should be in injected to libraries.

vishh:
LGTM

jiayingz:
Thanks a lot for the quick review!

PR #62, opened 2018-03-30
Update to gcr.io/cos-cloud/cos-gpu-installer:v20180329.

mindprince:
This updates the driver version to 390.46.
See GoogleCloudPlatform/cos-gpu-installer#10


mindprince:
This will unblock COS m66 testing.

PR #63, opened 2018-04-09
Addds k8s-app label in nvidia-driver-installer kube-system DaemonSet.

jiayingz:


mindprince:
Does both the DaemonSet and its Pods need this label?

Also, you need to update `nvidia-driver-installer/cos/...` as well.

jiayingz:
On Mon, Apr 9, 2018 at 4:16 PM Rohit Agarwal <notifications@github.com>
wrote:

> Does both the DaemonSet and its Pods need this label?
>
I actually don't know but found that seems the pattern used by other system
addons.

> Also, you need to update nvidia-driver-installer/cos/... as well.
>
Done.

PR #68, opened 2018-05-31
strip out vendor and ignore it, build deps at travis time

kodieGlosser:


mindprince:
Why?

kodieGlosser:
@mindprince The vendor directory should not be committed into github. If I try to do a `dep ensure -v` the code no longer compiles.

kodieGlosser:
I signed it!

kodieGlosser:
I signed it!

kodieGlosser:
@mindprince Does this make sense? This will basically pull in the code from the lock and update vendor directory. This is better than having a vendor directory pushed to the repo and greatly reduce the size of the project.

mindprince:
So, the pattern that we usually follow is to commit the vendor directory to ensure reproducible builds. This is what we do in Kubernetes and in all of our other projects. There are cases where the project is only delivering a library and in those cases, we sometime skip the committing the vendor directory and only commit the lock file but this is not one of those cases.

https://github.com/golang/dep/blob/master/docs/FAQ.md#should-i-commit-my-vendor-directory

> If I try to do a dep ensure -v the code no longer compiles.
The cloned copy of the code should already build, no need to run any dep commands unless you are modifying the dependencies.

I am going to close this.

PR #70, opened 2018-06-18
Ubuntu: Remove nouveau before installing nvidia drivers

AtzeDeVries:
By default (at least in my setup) ubuntu will load de nouveau kernel module. If the module is loaded the nvidia installer will fail. This PR contains a function to check on the module and remove it if needed.

### DIFF
--- a/nvidia-driver-installer/ubuntu/daemonset.yaml
+++ b/nvidia-driver-installer/ubuntu/daemonset.yaml
@@ -46,6 +46,9 @@ spec:
       - name: root-mount
         hostPath:
           path: /
+      - name: modules
+        hostPath:
+          path: /lib/modules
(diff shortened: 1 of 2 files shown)


AtzeDeVries:
I signed it!

mindprince:
Hey @AtzeDeVries,

I updated the ubuntu driver installer in #75 and #76 to handle restarts correctly.

The nouveau stuff unfortunately can't go in because we test this only on GCE/GKE and it's not needed there.

AtzeDeVries:
Ok, no problem. I've also changed my code a bit, also removed the nouveau part since it was not always working. I now updated the daemonset with an extra init container (before the driver install) that unloads nouveau.

PR #79, opened 2018-07-16
Add an example for training ResNet using GPU

yguo0905:
/assign @vishh 
/hold


mindprince:
Why create a new directory? There's already an existing examples directory.

vishh:
@mindprince we want to keep all the Next demo pieces in one directory for now.

PR #84, opened 2018-07-19
adding an image pre pull daemonset

vishh:
cc @yguo0905

### DIFF
--- a/demo/image-prepull-ds.yaml
+++ b/demo/image-prepull-ds.yaml
@@ -1,0 +1,86 @@
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: image-prepull
+spec:
+  selector:
+    matchLabels:
+      app: image-prepull
+  template:
+    metadata:
+      labels:
+        app: image-prepull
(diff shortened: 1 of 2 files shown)


demo/image-prepull-ds.yaml
```diff
@@ -0,0 +1,86 @@
...
+        effect: "NoExecute"
+      - operator: "Exists"
+        effect: "NoSchedule"
+      containers:
+      - args:
+        - -mode
+        - sleep
+        command:
+        - /drop/sleeper
+        image: gcr.io/vishnuk-cloud/tf-models-gpu:1.0
```
yguo0905: Since this image is general, we can publish it as `gcr.io/tensorflow/tpu-models-gpu:r1.9`. So users can try on their own.

Or maybe `gcr.io/tensorflow/models-gpu:r1.9` and `gcr.io/tensorflow/models-tpu:r1.9`?

We can change this later if we want.

vishh: +1. Do you want to open a CL for that?

PR #87, opened 2018-07-23
add minikube and gpu hyperparam examples

vishh:
Fixed all examples along the way


demo/tf-models-gpu/Dockerfile
```diff
@@ -12,10 +12,19 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM tensorflow/tensorflow:1.9.0-gpu
+FROM nvidia/cuda:9.0-cudnn7-runtime-ubuntu16.04
```
yguo0905: What's the reason of switching to using the cuda image and 1.8 model?

PR #89, opened 2018-08-10
Add daemonset that uses Canonical provided installer on GKE.

mindprince:


nvidia-driver-installer/ubuntu/daemonset-preloaded.yaml
```diff
@@ -0,0 +1,72 @@
...
+        hostPath:
+          path: /
+      initContainers:
+      - image: ubuntu-nvidia-installer:fixed
+        name: nvidia-driver-installer
+        resources:
+          requests:
+            cpu: 0.15
+        securityContext:
+          privileged: true
```
vishh: Do we need `privileged` or just a few capabilities would suffice?

mindprince: It's the same as the COS one for now. I think they do require some capabilities but I don't have a list.

vishh: Given we are starting from scratch, can we avoid using privileged?

mindprince: So, I tried `SYS_ADMIN` and `SYS_MODULE` but the installation failed. I will merge this for now so that the tests start using this and find all the capabilities that we need and send a follow up PR.

PR #91, opened 2018-08-11
CentOS driver installer

AtzeDeVries:
This PR has a CentOS 7 driver installer. Works the same as the Ubuntu installer. One difference is that the `umount /usr/bin` is not working as with Ubuntu. This is because in CentOS 7 `/bin` is a symbolic link to `/usr/bin` and the scripts starts bash so `/usr/bin/bash` is in use and unable to umount. A 'fixed' it by doing a lazy umount of `/usr/bin` (`umount -l /usr/bin`)

### DIFF
--- a/nvidia-driver-installer/centos/entrypoint.sh
+++ b/nvidia-driver-installer/centos/entrypoint.sh
@@ -1,0 +1,182 @@
+#!/bin/bash
+# Copyright 2017 Google Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


mindprince:
Sorry. This can't be merged. We don't have infrastructure setup to test that this installer remains working. (https://github.com/GoogleCloudPlatform/container-engine-accelerators/pull/54#issuecomment-367861500)

nvidia-driver-installer/centos/entrypoint.sh
```diff
@@ -0,0 +1,182 @@
...
+
+run_nvidia_installer() {
+  echo "Running Nvidia installer..."
+  pushd "${NVIDIA_INSTALL_DIR_CONTAINER}"
+  sh "${NVIDIA_INSTALLER_RUNFILE}" \
+    --utility-prefix="${NVIDIA_INSTALL_DIR_CONTAINER}" \
+    --opengl-prefix="${NVIDIA_INSTALL_DIR_CONTAINER}" \
+    --no-install-compat32-libs \
+    --log-file-name="${NVIDIA_INSTALL_DIR_CONTAINER}/nvidia-installer.log" \
+		--kernel-source-path=/usr/src/kernels/${KERNEL_VERSION} \
```
clkao: This should be /root/usr/src

PR #97, opened 2019-01-22
Parameterize kernel download and nvidia installer options.

clkao:
This allows reusing the entrypoint by overriding how the kernel is downloaded and where to look for it.

As an alternative to #91, you can use this for centos by setting the following extra env in the daemonset:

```
          - name: KERNEL_SOURCE_DOWNLOAD_COMMAND
            value: 'yum install -y kernel-devel-${KERNEL_VERSION}'
          - name: NVIDIA_INSTALLER_EXTRA_ARGS
            value: '--kernel-source-path=/usr/src/kernels/${KERNEL_VERSION}'

```

It is also possible to use this for airgapped install provided the kernel-devel package and .run file are available on the host:

```
          - name: NVIDIA_DRIVER_DOWNLOAD_URL
            value: file:///root/usr/src/NVIDIA-Linux-x86_64-384.130.run
          - name: KERNEL_SOURCE_DOWNLOAD_COMMAND
            value: 'echo'
          - name: NVIDIA_INSTALLER_EXTRA_ARGS
            value: '--kernel-source-path=/root/usr/src/kernels/${KERNEL_VERSION}'
```


clkao:
per https://github.com/GoogleCloudPlatform/container-engine-accelerators/pull/91#issuecomment-412379439 and #54, installers for additional distributions aren't likely to be merged due to maintainability.

If entrypoint.sh becomes common enough, we can rename ubuntu/ to generic/, and have additional Dockerfile.$distribution (that might not be tested and supported). This could be the case for the minikube installer too, as its entrypoint is largely the same.

PR #99, opened 2019-01-23
Updates Daemonset to Blanket Tolerance

ellis-bigelow:
Enables workload scheduling for tainted GPU workloads.

### DIFF
--- a/daemonset.yaml
+++ b/daemonset.yaml
@@ -42,9 +42,7 @@ spec:
               - key: cloud.google.com/gke-accelerator
                 operator: Exists
       tolerations:
-      - key: "nvidia.com/gpu"
-        effect: "NoSchedule"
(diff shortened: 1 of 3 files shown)


mindprince:
/lgtm

I will let @vishh merge this.

@thomas-riccardi This PR is for the driver installer daemonset (and not for the device plugin daemonset).

misterikkit:
@thomas-riccardi If you have a custom version of the nvidia drivers, maybe you could update the nodeAffinity of both DaemonSets to more specifically target the correct node pools.

thomas-riccardi:
@mindprince my bad, I assumed from the file paths that root `daemonset.yaml` was *not* for the driver.

@misterikkit thanks, I'll look into nodeAffinity.

PR #102, opened 2019-02-07
Sets LD_LIBRARY_PATH env in AllocateResponse so that gpu containers

jiayingz:
built with Cuda10 image can find nvidia libraries at proper place.

Tested manually with Cuda10 based vector-add:
https://github.com/kubernetes/kubernetes/pull/73823


vishh:
Did you test that it breaks without this change? If yes, please self-merge

jiayingz:
Yes.

houjun41544:
@jiayingz Have you made a PR to update the PATH variable?

jiayingz:
@houjun41544 Thanks for the reminding. Just merged the change. Did you run into any issue caused by the missing of PATH variable?

PR #103, opened 2019-02-15
Fix the crashing issue when nvidia.ko is already loaded.

### LINKED ISSUE #90: ubuntu driver install container crashed when nvidia.ko is already loaded.
In  https://github.com/GoogleCloudPlatform/container-engine-accelerators/blob/master/nvidia-driver-installer/ubuntu/entrypoint.sh there is a if statement to check if nvidia kernel module is loaded. It looks like this (line 139)
```
  if ! lsmod | grep -q -w 'nvidia'; then
    insmod "${NVIDIA_INSTALL_DIR_CONTAINER}/drivers/nvidia.ko"
  fi
  if ! lsmod | grep -q -w 'nvidia_uvm'; then
    insmod "${NVIDIA_INSTALL_DIR_CONTAINER}/drivers/nvidia-uvm.ko"
  fi
```
In `entrypoint.sh` at line 17 `set -o pipifail` is defined. This causes the check at line 139 to be always true and always tries to load the `nvidia.ko` even if it is already loaded. This causes the container to crash. 

A solution could be to remove `set -o pipefail`. I can do a PR if you want.

kaoet:
In the expression `! lsmod | grep -q -w nvidia`, the `-q` option
of `grep` makes it shortcircuit, which chokes the `lsmod` command,
causing it exit with non-zero.
Let's remove the `-q` option and replace it with `> /dev/null`.

Fixes #90

### DIFF
--- a/nvidia-driver-installer/minikube/entrypoint.sh
+++ b/nvidia-driver-installer/minikube/entrypoint.sh
@@ -174,13 +174,13 @@ run_nvidia_installer() {
 configure_cached_installation() {
   echo "Configuring cached driver installation..."
   update_container_ld_cache
-  if ! lsmod | grep -q -w 'nvidia'; then
+  if ! lsmod | grep -w 'nvidia' > /dev/null; then
     insmod "${NVIDIA_INSTALL_DIR_CONTAINER}/drivers/nvidia.ko"
   fi
-  if ! lsmod | grep -q -w 'nvidia_uvm'; then
+  if ! lsmod | grep -w 'nvidia_uvm' > /dev/null; then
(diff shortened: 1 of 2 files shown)


kaoet:
@mindprince Could you please take a look?

mindprince:
Thank you! Sorry it took a long time to get to this.

mindprince:
BTW, GKE now supports NVIDIA GPUs on Ubuntu image officially. Check out the docs for a faster installation experience: https://cloud.google.com/kubernetes-engine/docs/how-to/gpus#installing_drivers

PR #105, opened 2019-02-26
include xorg drivers

lsjostro:
xorg drivers is needed for remote rendering.

### DIFF
--- a/nvidia-driver-installer/ubuntu/Dockerfile
+++ b/nvidia-driver-installer/ubuntu/Dockerfile
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM ubuntu:16.04
+FROM ubuntu:bionic
 
 RUN apt-get update && \
     apt-get install -y kmod gcc make curl && \
--- a/nvidia-driver-installer/ubuntu/entrypoint.sh
+++ b/nvidia-driver-installer/ubuntu/entrypoint.sh
@@ -18,7 +18,7 @@ set -o pipefail
 set -u
 
 set -x
-NVIDIA_DRIVER_VERSION="${NVIDIA_DRIVER_VERSION:-384.111}"
+NVIDIA_DRIVER_VERSION="${NVIDIA_DRIVER_VERSION:-410.79}"
 NVIDIA_DRIVER_DOWNLOAD_URL_DEFAULT="https://us.download.nvidia.com/tesla/${NVIDIA_DRIVER_VERSION}/NVIDIA-Linux-x86_64-${NVIDIA_DRIVER_VERSION}.run"
 NVIDIA_DRIVER_DOWNLOAD_URL="${NVIDIA_DRIVER_DOWNLOAD_URL:-$NVIDIA_DRIVER_DOWNLOAD_URL_DEFAULT}"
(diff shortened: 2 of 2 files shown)


mindprince:
GKE officially supports installation of NVIDIA GPU drivers on Ubuntu now. The instructions to use that are here: https://cloud.google.com/kubernetes-engine/docs/how-to/gpus#installing_drivers

The officially supported version doesn't use this entrypoint.sh file (but uses a script present on the node image itself).

cc - @yguo0905 @jiayingz @vishh if it makes sense to add this to our official installation.

lsjostro:
yeah, and it doesn't include the X drivers unfortunately. Is the install script open sourced somewhere?
otherwise we need to roll our own.

vishh:
@yguo0905 can we get Canonical to include this change in their driver installer?

@mindprince if the Ubuntu installer is this repo is being widely used, we can accept this patch with a caveat that we do not intend to maintain it.

lsjostro:
here is the ubuntu nvidia deb package content (as you can see no xorg drivers included)
https://gist.github.com/lsjostro/7d2a468bcd7eca2865646f333517217a

lsjostro:
Canonical should btw open source their stuff? ;-)

jiayingz:
@lsjostro I saw you also updated the driver version. Is this to support the latest Cuda API?
Do you have a workload container that we may use for testing so that next time when we update the driver version again, we may run it to verify xorg applications still work?

lsjostro:
@jiayingz Sure, I'll publish a test example! And yes we need CUDA 10.

lsjostro:
@jiayingz https://github.com/EmbarkStudios/gke-accelerated-xorg-example
Test deployment manifest:
https://github.com/EmbarkStudios/gke-accelerated-xorg-example/blob/master/deployment.yaml

lsjostro:
any more thoughts on this?

jiayingz:
@lsjostro thanks a lot for the example test. Would your test work with the daemonset.yaml included in this directory? I manually built a ubuntu-nvidia-driver-installer with your patch applied and then run it with the latest ubuntu gke image. The driver installation finished successfully but when I launched your test deployment, it soon entered crashloop. Other than the changes included in this PR, do we also need any additional set up in the installer daemonset or any other configurations?

lsjostro:
@jiayingz The daemonset is exactly the same except for installer image from this PR.
But I'll try to reproduce. I'll get back to you.

bhack:
You can render headless with EGL and without GLX. 
https://github.com/NVIDIA/nvidia-docker/wiki/Frequently-Asked-Questions#is-opengl-supported

lsjostro:
I wish! Unity does not support EGL that I know unfortunately. 😩

bhack:
@lsjostro So these flags are still requiring GLX? https://answers.unity.com/questions/1488461/are-headless-batchmode-nographics-available-in-uni.html

lsjostro:
I believe so.. but I'm no unity Expert. I will double check with team. Cheers anyways.

potitoaghilar:
Is this still up? The result is a CrashLoopBack of xorg container due to no file found in path /dev/dri/card0. It seems GPUs are not accessible in this path

lsjostro:
Probably can close this now.

PR #109, opened 2019-04-15
Also ppopulates PATH env in AllocateResponse so that gpu containers

jiayingz:
can find gpu tools at right places.

Kudos to @RenaudWasTaken for the suggestion in
https://github.com/GoogleCloudPlatform/container-engine-accelerators/commit/65ea3c06db230e2061a03c251ae7908d5eb25a03#r32400515

### DIFF
--- a/pkg/gpu/nvidia/alpha_plugin.go
+++ b/pkg/gpu/nvidia/alpha_plugin.go
@@ -86,6 +86,7 @@ func (s *pluginServiceV1Alpha) Allocate(ctx context.Context, rqt *pluginapi.Allo
 	// in cuda10 docker ubuntu base images.
 	resp.Envs = make(map[string]string)
 	resp.Envs["LD_LIBRARY_PATH"] = "/usr/local/nvidia/lib:/usr/local/nvidia/lib64"
+	resp.Envs["PATH"] = "/usr/local/nvidia/bin:/usr/local/cuda/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
 	return resp, nil
 }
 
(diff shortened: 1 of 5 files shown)


pkg/gpu/nvidia/alpha_plugin.go
```diff
@@ -86,6 +86,7 @@ func (s *pluginServiceV1Alpha) Allocate(ctx context.Context, rqt *pluginapi.Allo
 	// in cuda10 docker ubuntu base images.
 	resp.Envs = make(map[string]string)
 	resp.Envs["LD_LIBRARY_PATH"] = "/usr/local/nvidia/lib:/usr/local/nvidia/lib64"
+	resp.Envs["PATH"] = "/usr/local/nvidia/bin:/usr/local/cuda/bin"
```
RenaudWasTaken: You are rewriting the path here, if someone set the path variable in his/her container this will cancel any changes the user made.

jiayingz: Right. I realized this issue and added the standard bin paths here. Lets discuss offline on the better long term solution.

RenaudWasTaken: `resp.Envs["PATH"] = "/usr/local/nvidia/bin:/usr/local/cuda/bin:$PATH"

jiayingz: I don't think $PATH would work with container runtime. I think container runtime will just overwrite with the last env it sees. Rethinking about this, it is probably not a good idea to overwrite popular env variables from device plugin. That logic is perhaps still better left to user image and container spec.

RenaudWasTaken: It will, the Env variables in CRI and OCI is represented as a list, not a map.

You can also get the intuition that it will work because of how this works in Dockerfiles, you can redeclare the PATH variable by using the previous `$PATH` declarations. It works pretty well like an overlay.

jiayingz:
/close

As mentioned above, overwriting $PATH env from device plugin may yield undesired behavior for users, so abandon this PR.

PR #110, opened 2019-04-19
Add periodic check for gpu devices installed after plugin server starts.

chardch:
New gpus that were installed after the device plugin server started were not being picked up by the plugin server and registered with the kubelet.

### DIFF
--- a/pkg/gpu/nvidia/manager.go
+++ b/pkg/gpu/nvidia/manager.go
@@ -32,12 +32,13 @@ import (
 const (
 	// All NVIDIA GPUs cards should be mounted with nvidiactl and nvidia-uvm
 	// If the driver installed correctly, these two devices will be there.
-	nvidiaCtlDevice = "/dev/nvidiactl"
-	nvidiaUVMDevice = "/dev/nvidia-uvm"
+	nvidiaCtlDevice = "nvidiactl"
+	nvidiaUVMDevice = "nvidia-uvm"
 	// Optional device.
-	nvidiaUVMToolsDevice = "/dev/nvidia-uvm-tools"
-	devDirectory         = "/dev"
-	nvidiaDeviceRE       = `^nvidia[0-9]*$`
+	nvidiaUVMToolsDevice      = "nvidia-uvm-tools"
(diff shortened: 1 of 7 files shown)


chardch:
I'm not sure why `make test` is failing. Locally, this works consistently. 
Screenshot from 2019-04-19 15-57-11

chardch:
The issue was fixed with making sure that the default devices were also created in addition to the gpu devices.

pkg/gpu/nvidia/manager.go
```diff
@@ -185,16 +232,37 @@ func (ngm *nvidiaGPUManager) Serve(pMountPath, kEndpoint, pluginEndpoint string)
 					glog.Infoln("device-plugin registered with the kubelet")
 				}
 
-				// This is checking if the plugin socket was deleted. If so,
-				// stop the grpc server and start the whole thing again.
+				// This is checking if the plugin socket was deleted
+				// and also if there are additional GPU devices installed.
+				// If so, stop the grpc server and start the whole thing again.
+				gpuCheck := time.NewTicker(10 * time.Second)
```
jiayingz: Perhaps define gpuCheckInterval and pluginSocketCheckInterval as const and use them in tests?

chardch:
Tested the behavior and it looks good. Used version 1.12.6-gke.10 with Ubuntu image and 8 gpus.

### Original state of fewer GPU devices registered:
gpu_incorrect_orig

The device plugin only makes one register request.
no_gpu_update_orig

### Using the change in this PR:

Reports the correct number of GPUs available: 8
correct_gpus

The device plugin makes two register requests when the new GPU devices are installed.
First request at 17:14:25, when there are only 2 devices installed and second at 17:14:38 when all 8 devices are installed.
gpus_updated

The GPU devices are not finished installing when the 
device_creation_on_updated_cluster

pkg/gpu/nvidia/manager.go
chardch: Thanks, added as consts. However it looks like the plugin socket interval is currently only used here and not in the tests?

PR #111, opened 2019-04-22
Revert "Sets LD_LIBRARY_PATH env in AllocateResponse so that gpu containers"

jiayingz:
Reverts GoogleCloudPlatform/container-engine-accelerators#102

As discussed in https://github.com/GoogleCloudPlatform/container-engine-accelerators/pull/109#discussion_r276040748 overwriting common envs like $PATH or $LD_LIBRARY_PATH from device plugin may yield undesired behavior for users, so reverting this PR.

Nvidia has already added these envs back for cuda 10.0 and 10.1 ubuntu base images:
https://gitlab.com/nvidia/cuda/commit/f9138b4eec5213c45303c8ea657e714dc6e60ae7
Customer images rebuilt with the latest ubuntu base image should be free of previous issues due to the missing of these env variables.

### DIFF
--- a/pkg/gpu/nvidia/alpha_plugin.go
+++ b/pkg/gpu/nvidia/alpha_plugin.go
@@ -82,10 +82,6 @@ func (s *pluginServiceV1Alpha) Allocate(ctx context.Context, rqt *pluginapi.Allo
 		HostPath:      s.ngm.hostPathPrefix,
 		ReadOnly:      true,
 	})
-	// Add LD_LIBRARY_PATH env to work around the compatibility issue
-	// in cuda10 docker ubuntu base images.
-	resp.Envs = make(map[string]string)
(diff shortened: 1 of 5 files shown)


chardch:
Looks good to me!
/lgtm

RenaudWasTaken:
I don't understand why you can't just append to the PATH and the LD_LIBRARY_PATH variable.
Setting the variables as follows will yield the best of both worlds:
`resp.Envs["PATH"] = "/usr/local/nvidia/bin:/usr/local/cuda/bin:$PATH"`

jiayingz:
@RenaudWasTaken as I learned, container runtime don't really interpret $xxx env variables. Suppose we have PATH="/bin" (from container spec) and PATH="/usr/bin:$PATH" (from device plugin) in ContainerRunOptions, they would get consolidated into a single setting PATH="/usr/bin:$PATH" by container runtime and how $PATH is interpreted would just depend on how that env is set in container image.

PR #112, opened 2019-05-17
Use versioned tags to be consistent with other addons in kubernetes

chardch:
Use human readable tags to be consistent with the other addons' images.
https://github.com/kubernetes/kubernetes/tree/master/cluster/addons


chardch:
+cc @jiayingz

PR #114, opened 2019-05-23
Add mount for Vulkan installable client driver to support Vulkan workloads

chardch:
Vulkan looks for a json file under /etc/vulkan/icd.d corresponding to the gpu driver being used.

### DIFF
--- a/cmd/nvidia_gpu/nvidia_gpu.go
+++ b/cmd/nvidia_gpu/nvidia_gpu.go
@@ -31,15 +31,20 @@ const (
 )
 
 var (
-	hostPathPrefix      = flag.String("host-path", "/home/kubernetes/bin/nvidia", "Path on the host that contains nvidia libraries. This will be mounted inside the container as '-container-path'")
-	containerPathPrefix = flag.String("container-path", "/usr/local/nvidia", "Path on the container that mounts '-host-path'")
-	pluginMountPath     = flag.String("plugin-directory", "/device-plugin", "The directory path to create plugin socket")
+	hostPathPrefix               = flag.String("host-path", "/home/kubernetes/bin/nvidia", "Path on the host that contains nvidia libraries. This will be mounted inside the container as '-container-path'")
+	containerPathPrefix          = flag.String("container-path", "/usr/local/nvidia", "Path on the container that mounts '-host-path'")
+	hostVulkanICDPathPrefix      = flag.String("host-vulkan-icd-path", "/home/kubernetes/bin/nvidia/vulkan/icd.d", "Path on the host that contains the Nvidia Vulkan installable client driver. This will be mounted inside the container as '-container-vulkan-icd-path'")
+	containerVulkanICDPathPrefix = flag.String("container-vulkan-icd-path", "/etc/vulkan/icd.d", "Path on the container that mounts '-host-vulkan-icd-path'")
+	pluginMountPath              = flag.String("plugin-directory", "/device-plugin", "The directory path to create plugin socket")
 )
 
 func main() {
 	flag.Parse()
 	glog.Infoln("device-plugin started")
-	ngm := gpumanager.NewNvidiaGPUManager(*hostPathPrefix, *containerPathPrefix, devDirectory)
+	mountPaths := []gpumanager.MountPath{
+		{HostPath: *hostPathPrefix, ContainerPath: *containerPathPrefix},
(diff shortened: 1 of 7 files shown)


cmd/nvidia_gpu/nvidia_gpu.go
```diff
@@ -31,15 +31,20 @@ const (
 )
 
 var (
-	hostPathPrefix      = flag.String("host-path", "/home/kubernetes/bin/nvidia", "Path on the host that contains nvidia libraries. This will be mounted inside the container as '-container-path'")
-	containerPathPrefix = flag.String("container-path", "/usr/local/nvidia", "Path on the container that mounts '-host-path'")
-	pluginMountPath     = flag.String("plugin-directory", "/device-plugin", "The directory path to create plugin socket")
+	hostPathPrefix               = flag.String("host-path", "/home/kubernetes/bin/nvidia", "Path on the host that contains nvidia libraries. This will be mounted inside the container as '-container-path'")
+	containerPathPrefix          = flag.String("container-path", "/usr/local/nvidia", "Path on the container that mounts '-host-path'")
+	hostVulkanICDPathPrefix      = flag.String("host-vulkan-icd-path", "/home/kubernetes/bin/nvidia", "Path on the host that contains the Nvidia Vulkan installable client driver. This will be mounted inside the container as '-container-vulkan-icd-path'")
+	containerVulkanICDPathPrefix = flag.String("container-vulkan-icd-path", "/etc/vulkan/icd.d", "Path on the container that mounts '-host-vulkan-icd-path'")
```
jiayingz: Do you know if vulkan/icd.d is also installed in order drivers? If the mount path doesn't exist, would a cuda container fail to start?

cmd/nvidia_gpu/nvidia_gpu.go
```diff
@@ -31,15 +31,20 @@ const (
 )
 
 var (
-	hostPathPrefix      = flag.String("host-path", "/home/kubernetes/bin/nvidia", "Path on the host that contains nvidia libraries. This will be mounted inside the container as '-container-path'")
-	containerPathPrefix = flag.String("container-path", "/usr/local/nvidia", "Path on the container that mounts '-host-path'")
-	pluginMountPath     = flag.String("plugin-directory", "/device-plugin", "The directory path to create plugin socket")
+	hostPathPrefix               = flag.String("host-path", "/home/kubernetes/bin/nvidia", "Path on the host that contains nvidia libraries. This will be mounted inside the container as '-container-path'")
+	containerPathPrefix          = flag.String("container-path", "/usr/local/nvidia", "Path on the container that mounts '-host-path'")
+	hostVulkanICDPathPrefix      = flag.String("host-vulkan-icd-path", "/home/kubernetes/bin/nvidia", "Path on the host that contains the Nvidia Vulkan installable client driver. This will be mounted inside the container as '-container-vulkan-icd-path'")
```
jiayingz: Should this be /home/kubernetes/bin/nvidia/vulkan/?

cmd/nvidia_gpu/nvidia_gpu.go
jiayingz: Should this be /etc/vulkan/icd.d?

chardch: Do you mean if this should be /etc/vulkan instead of /etc/vulkan/icd.d? I wasn't sure about this, but in the gpu installer container, I saw that the only thing under /etc/vulkan was icd.d/.

cmd/nvidia_gpu/nvidia_gpu.go
chardch: Thanks for catching.

cmd/nvidia_gpu/nvidia_gpu.go
chardch: I checked using the oldest supported supported driver, 390.46, and that creates vulkan/icd.d/nvidia_icd.json.

Also, I tested injecting a mount path that doesn't exist on the host and instead of gpu containers failing to start, the path was made on the host. The pod also wasn't privileged.

chardch: I tested using the non-existing path behavior using this device plugin image:  gcr.io/chardch-gke-dev/nvidia-gpu-device-plugin:1.0

jiayingz: I see. Thanks for the verification.

PR #118, opened 2019-07-16
Cache the driver .run file

mdaniel:
Currently the file unconditionally downloads the `.run` file, which weighs about 100MB, even if the file is already on disk from a previous failed run. This is needlessly slow and potentially a non-trivial amount of NAT gateway traffic when done cluster-wide

This change defaults to using the file if found, but offers the user an option through which they can opt-out of that behavior and always download the file even if found

I also took this opportunity to teach the entrypoint to remain in debug mode by request, which is infinitely handy for seeing what's going on, since one cannot use the "normal" trick of `command: ["bash", "-x", "/entrypoint.sh"]` because the script explicitly sets itself back to `+x`

### DIFF
--- a/nvidia-driver-installer/ubuntu/Dockerfile
+++ b/nvidia-driver-installer/ubuntu/Dockerfile
@@ -14,7 +14,8 @@
 
 FROM ubuntu:16.04
 
-RUN apt-get update && \
+RUN export DEBIAN_FRONTEND=noninteractive ;\
+    apt-get update && \
     apt-get install -y kmod gcc make curl && \
     rm -rf /var/lib/apt/lists/*
 
--- a/nvidia-driver-installer/ubuntu/entrypoint.sh
+++ b/nvidia-driver-installer/ubuntu/entrypoint.sh
@@ -27,7 +27,9 @@ NVIDIA_INSTALLER_RUNFILE="$(basename "${NVIDIA_DRIVER_DOWNLOAD_URL}")"
 ROOT_MOUNT_DIR="${ROOT_MOUNT_DIR:-/root}"
 CACHE_FILE="${NVIDIA_INSTALL_DIR_CONTAINER}/.cache"
 KERNEL_VERSION="$(uname -r)"
-set +x
+if [[ -z "${ENTRYPOINT_DEBUG:-}" ]]; then
+  set +x
+fi
 
 check_cached_version() {
   echo "Checking cached version"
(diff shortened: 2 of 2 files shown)


mdaniel:
Apologies, I don't know what that comment means (both overall and in the context of this PR); was it addressed to me?

chardch:
@mdaniel I had assumed this was being used on GKE. Please correct if otherwise. I wanted to note that there is official Ubuntu support on GKE: https://cloud.google.com/kubernetes-engine/docs/how-to/gpus#installing_drivers.

nvidia-driver-installer/ubuntu/entrypoint.sh
```diff
@@ -68,14 +71,15 @@ update_container_ld_cache() {
...
+  apt-get update
+  DEBIAN_FRONTEND=noninteractive apt-get install -y linux-headers-${KERNEL_VERSION}
   echo "Downloading kernel sources... DONE."
 }
 
 configure_nvidia_installation_dirs() {
   echo "Configuring installation directories..."
   mkdir -p "${NVIDIA_INSTALL_DIR_CONTAINER}"
-  pushd "${NVIDIA_INSTALL_DIR_CONTAINER}"
+  pushd    "${NVIDIA_INSTALL_DIR_CONTAINER}"
```
chardch: extra spaces?

nvidia-driver-installer/ubuntu/Dockerfile
```diff
@@ -14,7 +14,8 @@
 
 FROM ubuntu:16.04
 
-RUN apt-get update && \
+RUN export DEBIAN_FRONTEND=noninteractive ;\
```
chardch: Does DEBIAN_FRONTEND=noninteractive need to be set? These install with `-y`

nvidia-driver-installer/ubuntu/entrypoint.sh
```diff
@@ -113,6 +117,11 @@ configure_nvidia_installation_dirs() {
 download_nvidia_installer() {
   echo "Downloading Nvidia installer..."
   pushd "${NVIDIA_INSTALL_DIR_CONTAINER}"
+  if [[ -e "${NVIDIA_INSTALLER_RUNFILE}" && -z "${NVIDIA_INSTALLER_OVERWRITE:-}" ]]; then
+    echo "Using the existing \"${NVIDIA_INSTALLER_RUNFILE}\""
+    echo 'set NVIDIA_INSTALLER_OVERWRITE=1 to replace it with a fresh copy'
```
chardch: I think we should put this in a flag, instead of relying on the environment variable

PR #126, opened 2019-10-24
Revert mounting kernel source

chardch:
This breaks the nvidia-driver-installer on GKE versions prior to 1.14.


chardch:
Tested this on all currently supported GKE versions.

PR #128, opened 2019-11-22
Update apiVersion of cos daemonset to apps/v1 for branch stable.

lorqor:
apps/v1 has been supported since 1.12. This does not affect customers as
they're using branch master.


chardch:
Have you tested this out on all the supported gke versions?

PR #131, opened 2020-01-02
Add kustomization.yaml for cos

JohnPaton:
This will allow users to include this daemonset in their downstream builds by using
```yaml
resources:
- http://github.com/GoogleCloudPlatform/container-engine-accelerators.git//nvidia-driver-installer/cos/
```
in their own `kustomization.yaml` files

### DIFF
--- a/nvidia-driver-installer/cos/kustomization.yaml
+++ b/nvidia-driver-installer/cos/kustomization.yaml
@@ -1,0 +1,2 @@
+resources:
+- daemonset-preloaded.yaml


JohnPaton:
Pinging @chardch :)

chardch:
Thanks for the change @JohnPaton! Have you tested this on all supported GKE versions that support kustomization and the ones that don't?

JohnPaton:
Hi @chardch, kustomize is independent of GKE, it's just a tool that allows for the combination and patching of kubernetes manifests (like templating, but better). This will just allow users of kustomize to include this resource in their own manifests, and to patch it as needed (if desired).

JohnPaton:
@chardch is that clear enough?

pradvenkat (approved):
Thanks for the PR.

PR #136, opened 2020-04-04
Remove Cos-Tools from preloaded daemonset

bobbypage:
DO NOT SUBMIT


pradvenkat:
The fallback path in the cos-gpu-installer if fixed now. Should we just close this?

bobbypage:
Sounds good, thanks @pradvenkat

PR #143, opened 2020-08-03
Pin nvidia-driver-installer to specific version

pradvenkat:
Due to an update in the cos-gpu-installer image, the GPU OSS test started failing (see: https://github.com/kubernetes/kubernetes/issues/93187). The goal of the OSS test is not to test the driver installer itself, so I'd like to pin the driver installer version to a specific version, to avoid breaking tests when new driver installers are released.

### DIFF
--- a/daemonset.yaml
+++ b/daemonset.yaml
@@ -63,7 +63,7 @@ spec:
         hostPath:
           path: /
       initContainers:
-      - image: gcr.io/cos-cloud/cos-gpu-installer:latest
+      - image: gcr.io/cos-cloud/cos-gpu-installer@sha256:8d86a652759f80595cafed7d3dcde3dc53f57f9bc1e33b27bc3cfa7afea8d483  # 'v20200701'
         name: nvidia-driver-installer
         resources:
           requests:


daemonset.yaml
```diff
@@ -63,7 +63,7 @@ spec:
         hostPath:
           path: /
       initContainers:
-      - image: gcr.io/cos-cloud/cos-gpu-installer:latest
+      - image: gcr.io/cos-cloud/cos-gpu-installer@sha256:8d86a652759f80595cafed7d3dcde3dc53f57f9bc1e33b27bc3cfa7afea8d483
```
lorqor, line 66: Could you add a comment of the driver installer version (v20200701)?

pradvenkat: Done

pradvenkat:
> Would it make more sense to use tag instead of hash? So that if we make any change to the installer and want to test against that, we dont need to update the hash.

I'm trying to move away from using a tag so that the driver-installer image doesn't change unexpectedly. This makes it explicit which driver installer is being tested. If we want to test a newer driver installer, we can run the e2e tests locally with the updated hash, and then update daemonset.yaml in this repository - which is safer than updating the tag directly.

PR #144, opened 2020-08-11
Fix go module naming

pradvenkat:
In addition to making the name consistent, this makes it easier to test changes to `pkg/gpu/nvidia/`.  Without this change, builds used the packages from `vendor/`, but now they use the packages from this module directly.

Tested by successfully building device plugin image (`go build cmd/nvidia_gpu/nvidia_gpu.go`)


lorqor:
Thanks for the fix!

PR #149, opened 2020-10-17
Upgrade device plugin base image and golang version

pradvenkat:
1. We're running a very old version of golang (1.9). Upgrade to current stable version (1.15).
2. Since we're planning to pull in nvml into our device plugin, alpine base image no longer satisfies the required dependencies. Upgrade to debian. This increases the image size from ~7MB to ~50MB.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM golang:1.9-alpine as builder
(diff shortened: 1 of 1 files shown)


lorqor:
Would https://github.com/GoogleContainerTools/distroless satisfy the requirements for nvml?

pradvenkat:
Thanks for the great suggestion @lorqor! That is exactly what I was looking for. The image size is now down to ~15MB.  Updated PR.

PR #151, opened 2020-11-02
Prometheus endpoint in device plugin to expose GPU metrics

pradvenkat:
Summary of changes:

1.  Device plugin will expose a prometheus endpoint that exposes GPU metrics. This is behind a flag, and needs to be opted in.
2.  To get list of devices allocated to containers, device plugin will query kubelet's pod resources endpoint.
3.  Added dependency on nvml to query the GPU utilization metrics.


pkg/gpu/nvidia/metrics/metrics.go
```diff
@@ -0,0 +1,155 @@
...
+)
+
+var (
+	// DutyCycle reports the percent of time when the GPU was actively processing.
+	DutyCycle = promauto.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "duty_cycle",
+			Help: "Percent of time when the GPU was actively processing",
+		},
+		[]string{"namespace", "pod", "container", "make", "accelerator_id", "model"})
```
ruiwen-zhao, line 36: namespace, pod, container all seem to be high-cardinality fields, and it is against the best practice to store them in labels (https://prometheus.io/docs/practices/naming/#labels). Should we get rid of them?

pkg/gpu/nvidia/metrics/metrics.go
```diff
@@ -0,0 +1,155 @@
...
+		port:                port,
+		metricsEndpointPath: metricsEndpointPath,
+	}
+}
+
+// Start performs necessary initializations and starts the metric server.
+func (m *MetricServer) Start() error {
+	glog.Infoln("Starting metrics server")
+
+	err := nvml.Init()
```
ruiwen-zhao, line 82: Now we restrict the usage of nvml only within the metric reporting, but i guess in future we should move nvml to the core device plugin logic, i.e. the [ngm](https://github.com/GoogleCloudPlatform/container-engine-accelerators/blob/2651e63a6f7fc3682923219fa94321674c046559/pkg/gpu/nvidia/manager.go#L139)?

pkg/gpu/nvidia/metrics/metrics.go
```diff
@@ -0,0 +1,155 @@
...
+		}
+	}
+}
+
+func (m *MetricServer) updateMetrics(containerDevices map[ContainerID][]string) {
+	for container, devices := range containerDevices {
+		AcceleratorRequests.WithLabelValues(container.namespace, container.pod, container.container, gpuResourceName).Set(float64(len(devices)))
+
+		for _, device := range devices {
+			d, err := DeviceFromName(device)
```
ruiwen-zhao, line 132: why not call `DeviceUtilzation()` here?

ruiwen-zhao (commented):
Did we run any test with this change? Something like wget the prometheus port?

pkg/gpu/nvidia/metrics/devices.go
```diff
@@ -0,0 +1,146 @@
...
+
+	"github.com/NVIDIA/gpu-monitoring-tools/bindings/go/nvml"
+
+	"github.com/golang/glog"
+	"google.golang.org/grpc"
+	podresources "k8s.io/kubernetes/pkg/kubelet/apis/podresources/v1alpha1"
+)
+
+var (
+	socketPath      = "/var/lib/kubelet/pod-resources/kubelet.sock"
```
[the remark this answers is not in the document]
pradvenkat: Correct, I have a fork of the device plugin yaml I use for testing which mounts this path.

pkg/gpu/nvidia/metrics/devices.go
```diff
@@ -0,0 +1,146 @@
...
+				}
+			}
+		}
+	}
+
+	return containerDevices, nil
+}
+
+// DiscoverGPUDevices discovers GPUs attached to the node, and updates `gpuDevices` map.
+func DiscoverGPUDevices() error {
```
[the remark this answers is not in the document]
pradvenkat: Yup exactly. Right now the goal is to keep all new changes isolated. I think longer term, the device plugin could just rely on nvml to discover devices as well. But that would be a separate change.

pkg/gpu/nvidia/metrics/metrics.go
pradvenkat: Actually on a single node, we can have at most 16 GPUs. Each GPU can be allocated to only one container at a time. So the cardinality is not that high.

The reason we need these labels is because these metrics are published to the k8s_container resource (https://cloud.google.com/monitoring/api/resources#tag_k8s_container) in stack driver. namespace, pod and container fields will be derived from these labels.

pkg/gpu/nvidia/metrics/metrics.go
pradvenkat: Completely agree with the future direction. Right now, completely isolating the new functionality behind a kill switch seems to be the safest option.

pkg/gpu/nvidia/metrics/metrics.go
pradvenkat: We anyway need to get the device from the device name so that we can reference other fields like UUID and Model. At that point, DeviceUtilzation() function doesn't add any value, since we can directly call d.Status() once we have the device pointer.  I think we can just remove the DeviceUtilzation() method, since it's not needed.

pradvenkat:
For testing, I created a new device plugin image with the enable metrics flag. I forked the device plugin yaml to use the new image, and also mounted the kubelet socket dir and set the LD_LIBRARY_PATH env variable (to discover nvidia shared libs). The I deployed a GPU workload, and queried for the metrics:

```
$ kubectl --namespace=kube-system port-forward nvidia-gpu-device-plugin-<...>  2112:2112

$ curl http://127.0.0.1:2112/metrics | grep duty                                                                                                                                                                                                                                                   

# TYPE duty_cycle gauge                                                                                                                                                                                                                                                                                                                          
duty_cycle{accelerator_id="GPU-1592a165-0721-29ee-004e-b02e5b1cdaa1",container="gpu-perf",make="nvidia",model="A100-SXM4-40GB",namespace="default",pod="gpu-perf"} 13
```

PR #155, opened 2020-11-11
Device plugin support GPU partitions (MIG prototype)

pradvenkat:


pkg/gpu/nvidia/mig/mig.go
```diff
@@ -0,0 +1,357 @@
...
+	if !ok {
+		return []pluginapi.DeviceSpec{}, fmt.Errorf("invalid allocation request with non-existing GPU partition: %s", deviceID)
+	}
+
+	return deviceSpecs, nil
+}
+
+// Start method performs the necessary initializations and starts the mig.DeviceManager.
+func (d *DeviceManager) Start() error {
+	if err := nvml.Init(); err != nil {
```
ruiwen-zhao, line 109: is it ok to init 2 nvmls? coz there is another nvml running for monitoring GPUs

pkg/gpu/nvidia/mig/mig.go
```diff
@@ -0,0 +1,357 @@
...
+func createDeviceNode(path string, major, minor uint32) error {
+	if err := syscall.Mknod(path, syscall.S_IFCHR|uint32(os.FileMode(0444)), int(unix.Mkdev(major, minor))); err != nil {
+		return fmt.Errorf("Failed to create devive node %s: %v", path, err)
+	}
+	return nil
+}
+
+func deleteAllGPUPartitions() error {
+	// TODO: Use NVML to delete GPU partitions.
+	// Currently we are unable to use NVML to delete GPU partitions from inside the device plugin container, so using nvidia-smi for now.
```
ruiwen-zhao, line 341: why are we unable to do it within device plugin?

pkg/gpu/nvidia/mig/mig.go
```diff
@@ -0,0 +1,357 @@
...
+	return nvml.Shutdown()
+}
+
+// ConfigureGPUPartitions accepts a string describing the GPU partition type, and partitions all the GPUs on the node as specified.
+func (d *DeviceManager) ConfigureGPUPartitions(partition string) error {
+	if err := d.enableMIGMode(); err != nil {
+		return fmt.Errorf("failed to configure GPU partitions: %v", err)
+	}
+
+	if err := deleteAllGPUPartitions(); err != nil {
```
ruiwen-zhao, line 140: is this needed for device plugin restarts, where we need to clean up the existing partitions first?

pkg/gpu/nvidia/mig/mig.go
```diff
@@ -0,0 +1,357 @@
...
+		rebootNeeded = true
+
+		_, err = device.SetMigMode(nvml.DEVICE_MIG_ENABLE)
+		if err != nil {
+			return fmt.Errorf("could not enable mig mode on GPU %s: %v", device.Path, err)
+		}
+	}
+
+	if rebootNeeded {
+		glog.Infof("Node needs to rebooted after turning on MIG mode on GPUs")
```
ruiwen-zhao, line 181: nit: "needs to be rebooted"

pkg/gpu/nvidia/mig/mig.go
```diff
@@ -0,0 +1,357 @@
...
+		err := rebootNode()
```
ruiwen-zhao, line 182: this doesnt seem to be blocking. So it is possible that `enableMIGMode()` returns with no error but the node is not actually in MIG mode? Should we return an error here telling the caller that the node is not yet in MIG mode?

pkg/gpu/nvidia/mig/mig.go
```diff
@@ -0,0 +1,357 @@
...
+					Permissions:   "mrw",
+				},
+			}
+
+		}
+	}
+	return nil
+}
+
+func createDeviceNode(path string, major, minor uint32) error {
```
ruiwen-zhao, line 332: nit: the function name `createDeviceNode()` is a bit confusing because `CreateDeviceNodes()` seems to be doing way more work than this one does.

pkg/gpu/nvidia/mig/mig.go
pradvenkat: I think it should be fin. All it does is call `dlopen` on libnvml.so. If there are multiple dlopen() calls, we still get the same handle. Eventually we can move the initialization to a common place, but for now just keeping the features isolated.

pkg/gpu/nvidia/mig/mig.go
pradvenkat: Good point. Ideally if the GPU is already partitioned the way we expect, then we should just skip that step. I'll update this.

pkg/gpu/nvidia/mig/mig.go
pradvenkat: I'm not sure actually. Trying to delete a GPU partition just results in invalid argument error. We may need to look deeper into it.

evol262:
@pradvenkat Is anything holding up the merge here other than a rebase?

pradvenkat:
> @pradvenkat Is anything holding up the merge here other than a rebase?

I'm planning to the change the approach a little bit. I want to move the MIG device creation logic out of the device plugin, and have the device plugin just focus on discovering GPU partitions.

pradvenkat:
As mentioned earlier, I'm going to close this PR in favor of https://github.com/GoogleCloudPlatform/container-engine-accelerators/pull/170 and https://github.com/GoogleCloudPlatform/container-engine-accelerators/pull/171.

PR #156, opened 2020-11-13
Fix type

AlexeyPerevalov:
it affects build


ruiwen-zhao:
hmm interesting, why didnt we find this before? Does this only affect build when the Topology related stuff is added?

AlexeyPerevalov:
> hmm interesting, why didnt we find this before? Does this only affect build when the Topology related stuff is added?

no it wasn't root cause. 
I think make vet is not working properly in this repo, I can take a look into it.
Also when I tried nvidia-device-plugin localy on ubuntu 20.04, I faced with several troubles:
1. ./nvidia-driver-installer/ubuntu is for 16.04, it was problem with downloading actual kernel source
2. when docker image was changed to 20.04 I faced with problem in configure_nvidia_installation_dirs it creates directory  /usr/local/nvidia and then bin & bin-workdir, at mount -t overlay it fails, since mount expects read-only file system in my case.

PR #157, opened 2020-11-23
Device plugin metrics improvements

pradvenkat:
Couple of improvements:
1. Stop reporting metrics for containers after they've terminated (periodically reset all metrics).
2. Report memory_used and memory_total in bytes.

### DIFF
--- a/cmd/nvidia_gpu/nvidia_gpu.go
+++ b/cmd/nvidia_gpu/nvidia_gpu.go
@@ -39,7 +39,7 @@ var (
 	pluginMountPath                = flag.String("plugin-directory", "/device-plugin", "The directory path to create plugin socket")
 	enableContainerGPUMetrics      = flag.Bool("enable-container-gpu-metrics", false, "If true, the device plugin will expose GPU metrics for containers with allocated GPU")
 	gpuMetricsPort                 = flag.Int("gpu-metrics-port", 2112, "POrt on which GPU metrics for containers are exposed")
-	gpuMetricsCollectionIntervalMs = flag.Int("gpu-metrics-collection-interval", 2000, "Colection interval (in milli seconds) for container GPU metrics")
+	gpuMetricsCollectionIntervalMs = flag.Int("gpu-metrics-collection-interval", 30000, "Colection interval (in milli seconds) for container GPU metrics")
 )
 
 func main() {
--- a/pkg/gpu/nvidia/metrics/metrics.go
+++ b/pkg/gpu/nvidia/metrics/metrics.go
@@ -60,18 +60,22 @@ var (
 		[]string{"namespace", "pod", "container", "resource_name"})
 )
 
+const metricsResetInterval = time.Minute
+
 // MetricServer exposes GPU metrics for all containers in prometheus format on the specified port.
 type MetricServer struct {
-	collectionInterval  int
-	port                int
-	metricsEndpointPath string
+	collectionInterval   int
+	port                 int
+	metricsEndpointPath  string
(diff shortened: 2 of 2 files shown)


cmd/nvidia_gpu/nvidia_gpu.go
```diff
@@ -39,7 +39,7 @@ var (
 	pluginMountPath                = flag.String("plugin-directory", "/device-plugin", "The directory path to create plugin socket")
 	enableContainerGPUMetrics      = flag.Bool("enable-container-gpu-metrics", false, "If true, the device plugin will expose GPU metrics for containers with allocated GPU")
 	gpuMetricsPort                 = flag.Int("gpu-metrics-port", 2112, "POrt on which GPU metrics for containers are exposed")
-	gpuMetricsCollectionIntervalMs = flag.Int("gpu-metrics-collection-interval", 2000, "Colection interval (in milli seconds) for container GPU metrics")
+	gpuMetricsCollectionIntervalMs = flag.Int("gpu-metrics-collection-interval", 30000, "Colection interval (in milli seconds) for container GPU metrics")
```
ruiwen-zhao, line 42: Why do we change the interval here?

pkg/gpu/nvidia/metrics/metrics.go
```diff
@@ -148,12 +154,23 @@ func (m *MetricServer) updateMetrics(containerDevices map[ContainerID][]string)
...
+			MemoryTotal.WithLabelValues(container.namespace, container.pod, container.container, "nvidia", d.UUID, *d.Model).Set(float64(*d.Memory) * 1024 * 1024)       // memory reported in bytes
+			MemoryUsed.WithLabelValues(container.namespace, container.pod, container.container, "nvidia", d.UUID, *d.Model).Set(float64(*utilInfo.Memory) * 1024 * 1024) // memory reported in bytes
 		}
 	}
 }
 
+func (m *MetricServer) resetMetricsIfNeeded() {
+	if time.Now().After(m.lastMetricsResetTime.Add(metricsResetInterval)) {
+		AcceleratorRequests.Reset()
+		DutyCycle.Reset()
```
ruiwen-zhao, line 166: I don't really understand what the `Reset` does here. The [prometheus doc](https://sourcegraph.com/github.com/kubernetes/component-base/-/blob/metrics/gauge.go#L165:20&tab=references) says it deletes all the metrics. So will we lose metrics by doing this? And why this will prevent terminated containers from emitting metrics?

cmd/nvidia_gpu/nvidia_gpu.go
pradvenkat: Just to reduce the amount of unnecessary work the device plugin needs to do. We plan to collect the GPU metrics only once a minute,  so having the device plugin collect the metrics every 30 seconds should be sufficient.

pkg/gpu/nvidia/metrics/metrics.go
pradvenkat: The idea is that we periodically reset all the metrics, but continue to publish the metrics for containers that are still running. This method is called at the beginning of `updateMetrics()`. So we clear the metrics followed by immediately discovering all running containers, and publishing those metrics again.  Since the number of GPU containers on the node should be very small (<= 16), this should approach should work well, while keeping the code simple.

We could do something more sophisticated like keeping track of running containers, and deleting the corresponding metrics as they terminate - but it will add lot more complexity.

ruiwen-zhao (approved):

PR #159, opened 2020-12-01
Fix memory used metric reported by device plugin

pradvenkat:


pkg/gpu/nvidia/metrics/metrics.go
```diff
@@ -155,7 +155,7 @@ func (m *MetricServer) updateMetrics(containerDevices map[ContainerID][]string)
 
 			DutyCycle.WithLabelValues(container.namespace, container.pod, container.container, "nvidia", d.UUID, *d.Model).Set(float64(dutyCycle))
 			MemoryTotal.WithLabelValues(container.namespace, container.pod, container.container, "nvidia", d.UUID, *d.Model).Set(float64(*d.Memory) * 1024 * 1024)       // memory reported in bytes
-			MemoryUsed.WithLabelValues(container.namespace, container.pod, container.container, "nvidia", d.UUID, *d.Model).Set(float64(*utilInfo.Memory) * 1024 * 1024) // memory reported in bytes
+			MemoryUsed.WithLabelValues(container.namespace, container.pod, container.container, "nvidia", d.UUID, *d.Model).Set(float64(*mem.Global.Used) * 1024 * 1024) // memory reported in bytes
```
ruiwen-zhao, line 158: what's the difference between these two?

ruiwen-zhao (approved):

pkg/gpu/nvidia/metrics/metrics.go
pradvenkat: With this change, we report the actual memory used. The earlier metric refers to the percentage of time where the device mem was read or written to (https://github.com/NVIDIA/gpu-monitoring-tools/blob/master/bindings/go/nvml/nvml.h#L1580).

PR #164, opened 2020-12-12
Move NVML initialization and shutdown to main function

ruiwen-zhao:
Moving NVML initialization and shutdown to main function, so that nvidia manager and metric server can both access to NVML and avoid re-initializing it.

### DIFF
--- a/cmd/nvidia_gpu/nvidia_gpu.go
+++ b/cmd/nvidia_gpu/nvidia_gpu.go
@@ -21,6 +21,7 @@ import (
 
 	gpumanager "github.com/GoogleCloudPlatform/container-engine-accelerators/pkg/gpu/nvidia"
 	"github.com/GoogleCloudPlatform/container-engine-accelerators/pkg/gpu/nvidia/metrics"
+	"github.com/NVIDIA/gpu-monitoring-tools/bindings/go/nvml"
 	"github.com/golang/glog"
 )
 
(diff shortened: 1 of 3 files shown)


pkg/gpu/nvidia/nvml/nvml_manager.go
```diff
@@ -0,0 +1,48 @@
+package nvmlmanager
```
pradvenkat, line 1: Keeping with the go convention of package name being the last element of package path, can we name this `package nvml`?

cmd/nvidia_gpu/nvidia_gpu.go
```diff
@@ -48,6 +49,10 @@ func main() {
 	mountPaths := []gpumanager.MountPath{
 		{HostPath: *hostPathPrefix, ContainerPath: *containerPathPrefix},
 		{HostPath: *hostVulkanICDPathPrefix, ContainerPath: *containerVulkanICDPathPrefix}}
+
+	nvmlManager := nvmlmanager.GetInstance()
```
pradvenkat, line 53: Why not consolidate nvml lifecycle management in the main function itself? That way we do not have to pass the nvmlmanager object around.  It will also prevent any object from incorrectly calling `nvmlmanager.Shutdown()` while some other object may still be using it.

What if we re-write this main function to make the steps more clear:
1.  Create a new method to check if drivers are installed, and keep looping till we detect drivers are installed.
2.  Call nvml.Init() and defer nvml.Shutdown().
3.  Start gpumanager
4.  Start metrics
5.  Start any new components

What do you think?

pradvenkat (commented):
Can you remove `cmd/nvidia_gpu/nvidia_gpu` from the commit?

cmd/nvidia_gpu/nvidia_gpu.go
ruiwen-zhao: yeah i guess that will be more clear and save us the trouble of introducing the manager. Will do it

robertdavidsmith (commented):
Note accidental commit of binary file cmd/nvidia_gpu/nvidia_gpu

pradvenkat (approved):

PR #165, opened 2020-12-14
Report NUMA node for use by K8s TopologyManager

robertdavidsmith:
This PR makes the device plugin report GPU NUMA node for use by K8s TopologyManager.

First it uses NVML to map device id to pci bus id. Then it uses the pci bus id to find the relevant numa_node file under /sys, parses this file and returns the NUMA node to k8s via protobuf.

This feature is off by default, but can be enabled via the -topology command line option.


robertdavidsmith:
@googlebot I signed it!

PR #171, opened 2021-01-08
Simple tool to partition GPUs on a node

pradvenkat:


partition_gpu/partition_gpu.yaml
```diff
@@ -0,0 +1,78 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This daemonset deploys the GPU partitioner on all GPU nodes and partitions
+# the GPUs as defined in the GPU config file.
+
+apiVersion: apps/v1
+kind: DaemonSet
```
ruiwen-zhao, line 19: this eventually will become an init container of the device plugin daemonset right?

pradvenkat: Yup, that's the plan. For now, since this can be used independently, I created a daemonset for it.

PR #175, opened 2021-01-15
Catch critical Xid errors and mark GPUs unhealthy

ruiwen-zhao:
# What's changed
1. Added a `GPUHealthChecker` component which registers NVML events and listens to them.
2.  When an NVML event pops up, `GPUHealthChecker` will pass the device with updated healthness to `ListAndWatch` through a channel. `ListAndWatch` will update `ngm.devices` and send the latest device status to kubelet.
3. All `CheckHealth` logic will only be triggered when `--enable-health-monitoring` flag is used. This PR will be a no-op if the flag is not used, although the two `ListAndWatch` methods are slightly refactored.

# Test
* Added a piece of code to device manager that will periodically flap the health of the GPU (something like `ngm.devices['nvidia0'] = 'Unhealthy'`), deployed it to a cluster, and observed the allocatable fields in the node yaml. I can see that the number of allocatable GPUs changes accordingly. 
* Will need some further manual testing to test the NVML event part.

### DIFF
--- a/cmd/nvidia_gpu/nvidia_gpu.go
+++ b/cmd/nvidia_gpu/nvidia_gpu.go
@@ -22,6 +22,7 @@ import (
 	"time"
 
 	gpumanager "github.com/GoogleCloudPlatform/container-engine-accelerators/pkg/gpu/nvidia"
+	healthcheck "github.com/GoogleCloudPlatform/container-engine-accelerators/pkg/gpu/nvidia/health_check"
 	"github.com/GoogleCloudPlatform/container-engine-accelerators/pkg/gpu/nvidia/metrics"
 	"github.com/NVIDIA/gpu-monitoring-tools/bindings/go/nvml"
 	"github.com/golang/glog"
@@ -41,6 +42,7 @@ var (
 	containerVulkanICDPathPrefix   = flag.String("container-vulkan-icd-path", "/etc/vulkan/icd.d", "Path on the container that mounts '-host-vulkan-icd-path'")
 	pluginMountPath                = flag.String("plugin-directory", "/device-plugin", "The directory path to create plugin socket")
 	enableContainerGPUMetrics      = flag.Bool("enable-container-gpu-metrics", false, "If true, the device plugin will expose GPU metrics for containers with allocated GPU")
+	enableHealthMonitoring         = flag.Bool("enable-health-monitoring", false, "If true, the device plugin will detect critical Xid errors and mark the GPUs unallocatable")
 	gpuMetricsPort                 = flag.Int("gpu-metrics-port", 2112, "POrt on which GPU metrics for containers are exposed")
 	gpuMetricsCollectionIntervalMs = flag.Int("gpu-metrics-collection-interval", 30000, "Colection interval (in milli seconds) for container GPU metrics")
 	gpuConfigFile                  = flag.String("gpu-config", "/etc/nvidia/gpu_config.json", "File with GPU configurations for device plugin")
@@ -110,6 +112,15 @@ func main() {
 			return
 		}
 		defer metricServer.Stop()
+	}
+
+	if *enableHealthMonitoring {
+		hc := healthcheck.NewGPUHealthChecker(ngm.ListDevices(), ngm.Health)
+		if err := hc.Start(); err != nil {
+			glog.Infof("Failed to start GPU Health Checker: %v", err)
+			return
+		}
+		defer hc.Stop()
 	}
 
 	ngm.Serve(*pluginMountPath, kubeletEndpoint, fmt.Sprintf("%s-%d.sock", pluginEndpointPrefix, time.Now().Unix()))
--- a/pkg/gpu/nvidia/alpha_plugin.go
+++ b/pkg/gpu/nvidia/alpha_plugin.go
@@ -33,22 +33,18 @@ type pluginServiceV1Alpha struct {
 
 func (s *pluginServiceV1Alpha) ListAndWatch(emtpy *pluginapi.Empty, stream pluginapi.DevicePlugin_ListAndWatchServer) error {
 	glog.Infoln("device-plugin: ListAndWatch start")
-	changed := true
+	if err := s.sendDevices(stream); err != nil {
+		return err
+	}
 	for {
-		if changed {
-			resp := new(pluginapi.ListAndWatchResponse)
-			for _, dev := range s.ngm.devices {
-				resp.Devices = append(resp.Devices, &pluginapi.Device{ID: dev.ID, Health: dev.Health})
-			}
-			glog.Infof("ListAndWatch: send devices %v\n", resp)
-			if err := stream.Send(resp); err != nil {
-				glog.Errorf("device-plugin: cannot update device states: %v\n", err)
-				s.ngm.grpcServer.Stop()
+		select {
+		case d := <-s.ngm.Health:
+			glog.Infof("device-plugin: %s device marked as %s", d.ID, d.Health)
+			s.ngm.devices[d.ID] = d
+			if err := s.sendDevices(stream); err != nil {
 				return err
 			}
 		}
-		time.Sleep(5 * time.Second)
-		changed = s.ngm.CheckDeviceStates()
 	}
 }
 
@@ -112,6 +108,20 @@ func RegisterWithKubelet(kubeletEndpoint, pluginEndpoint, resourceName string) e
 
 	if _, err = client.Register(context.Background(), request); err != nil {
 		return fmt.Errorf("device-plugin: cannot register to kubelet service: %v", err)
+	}
+	return nil
+}
+
+func (s *pluginServiceV1Alpha) sendDevices(stream pluginapi.DevicePlugin_ListAndWatchServer) error {
+	resp := new(pluginapi.ListAndWatchResponse)
+	for _, dev := range s.ngm.ListDevices() {
+		resp.Devices = append(resp.Devices, &pluginapi.Device{ID: dev.ID, Health: dev.Health})
+	}
+	glog.Infof("ListAndWatch: send devices %v\n", resp)
+	if err := stream.Send(resp); err != nil {
+		glog.Errorf("device-plugin: cannot update device states: %v\n", err)
+		s.ngm.grpcServer.Stop()
+		return err
 	}
 	return nil
 }
--- a/pkg/gpu/nvidia/health_check/health_checker.go
+++ b/pkg/gpu/nvidia/health_check/health_checker.go
@@ -1,0 +1,168 @@
+// Copyright 2017 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package healthcheck
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/GoogleCloudPlatform/container-engine-accelerators/pkg/gpu/nvidia/util"
+	"github.com/NVIDIA/gpu-monitoring-tools/bindings/go/nvml"
+	"github.com/golang/glog"
+
+	pluginapi "k8s.io/kubelet/pkg/apis/deviceplugin/v1beta1"
+)
+
+// GPUHealthChecker checks the health of nvidia GPUs. Note that with the current
+// device naming pattern in device manager, GPUHealthChecker will not work with
+// MIG devices.
+type GPUHealthChecker struct {
+	devices     map[string]pluginapi.Device
+	nvmlDevices map[string]*nvml.Device
+	health      chan pluginapi.Device
+	eventSet    nvml.EventSet
+	stop        chan bool
+}
+
+// NewGPUHealthChecker returns a GPUHealthChecker object for a given device name
+func NewGPUHealthChecker(devices map[string]pluginapi.Device, health chan pluginapi.Device) *GPUHealthChecker {
+	return &GPUHealthChecker{
+		devices:     devices,
+		nvmlDevices: make(map[string]*nvml.Device),
+		health:      health,
+		stop:        make(chan bool),
+	}
+}
+
+// Start registers NVML events and starts listening to them
+func (hc *GPUHealthChecker) Start() error {
+	glog.Info("Starting GPU Health Checker")
+
+	// Building mapping between device ID and their nvml represetation
+	count, err := nvml.GetDeviceCount()
+	if err != nil {
+		return fmt.Errorf("failed to get device count: %s", err)
+	}
+
+	glog.Infof("Found %d GPU devices", count)
+	for i := uint(0); i < count; i++ {
+		device, err := nvml.NewDeviceLite(i)
+		if err != nil {
+			return fmt.Errorf("failed to read device with index %d: %v", i, err)
+		}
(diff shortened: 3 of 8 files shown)


pkg/gpu/nvidia/manager.go
```diff
@@ -339,5 +448,18 @@ func (ngm *nvidiaGPUManager) Stop() error {
 	}
 	ngm.stop <- true
 	<-ngm.stop
+	if ngm.healthCheckRunning {
+		ngm.healthCheckStop <- true
+		<-ngm.healthCheckStop
```
pradvenkat, line 453: What do we need this for?

pkg/gpu/nvidia/manager.go
```diff
@@ -339,5 +448,18 @@ func (ngm *nvidiaGPUManager) Stop() error {
...
+	}
```
pradvenkat, line 454: Should we close ngm.health channel as well?

pkg/gpu/nvidia/manager.go
```diff
@@ -235,6 +244,106 @@ func (ngm *nvidiaGPUManager) CheckDeviceStates() bool {
 	return changed
 }
 
+// CheckHealth registers event with NVML to detect GPU critical errors
+func (ngm *nvidiaGPUManager) CheckHealth() error {
```
pradvenkat, line 248: To prevent the manager package from bloating too much, would it be better to move the health check functionality to a separate package? The device manager and health checker just need a channel to communicate. Separating health check piece out would make it easier to unit test, and read/make changes in the future.

pradvenkat (commented):
Can you help me understand how the health check will work when MIG mode is enabled, and GPUs are partitioned?

cmd/nvidia_gpu/nvidia_gpu.go
```diff
@@ -112,5 +113,13 @@ func main() {
 		defer metricServer.Stop()
 	}
 
+	if *enableHealthMonitoring {
+		go func() {
+			if err := ngm.CheckHealth(); err != nil {
+				glog.Errorf("CheckHealth exited with error: %v. Device plugin will keep running.", err)
```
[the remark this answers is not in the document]
ruiwen-zhao: I was just trying to make this new feature less disruptive. But yeah I guess it does make more sense to restart device plugin when the health check fails to start. Will make the change.

A flag to ignore errors to to me is basically a synonym of not enabling health monitoring.

pkg/gpu/nvidia/alpha_plugin.go
```diff
@@ -33,22 +33,17 @@ type pluginServiceV1Alpha struct {
...
-			resp := new(pluginapi.ListAndWatchResponse)
-			for _, dev := range s.ngm.devices {
-				resp.Devices = append(resp.Devices, &pluginapi.Device{ID: dev.ID, Health: dev.Health})
-			}
-			glog.Infof("ListAndWatch: send devices %v\n", resp)
-			if err := stream.Send(resp); err != nil {
-				glog.Errorf("device-plugin: cannot update device states: %v\n", err)
-				s.ngm.grpcServer.Stop()
+		select {
+		case d := <-s.ngm.health:
```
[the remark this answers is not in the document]
ruiwen-zhao: The channel should never be closed (no one will close it), and it will only have values when a device goes unhealthy, so there should not be any unnecessary `sendDevices` calls.

pkg/gpu/nvidia/manager.go
ruiwen-zhao: I was thinking about this as well. But CheckHealth basically deals with the device status, so logically it should be part of device manager.
If we separate it out, we will need one channel between `CheckHealth` and device manager so that device manager can update the status, and another channel between device manager and `ListAndWatch`. And that seems a bit too complicated. Thoughts?

pkg/gpu/nvidia/manager.go
```diff
@@ -235,6 +244,106 @@ func (ngm *nvidiaGPUManager) CheckDeviceStates() bool {
...
+			continue
+		}
+		uuid := nd.UUID
+		gpu, _, _, err := nvml.ParseMigDeviceUUID(uuid)
+		if err != nil {
+			gpu = uuid
+		}
+		err = nvml.RegisterEventForDevice(eventSet, nvml.XidCriticalError, gpu)
+		glog.Infof("Registering device %v. UUID: %s", d.ID, uuid)
+		if err != nil && strings.HasSuffix(err.Error(), "Not Supported") {
```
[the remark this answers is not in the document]
ruiwen-zhao: good catch. I will log an error and exit device plugin if it is some other error.

pkg/gpu/nvidia/manager.go
```diff
@@ -235,6 +244,106 @@ func (ngm *nvidiaGPUManager) CheckDeviceStates() bool {
...
+		select {
+		case <-ngm.healthCheckStop:
+			ngm.healthCheckRunning = false
+			close(ngm.healthCheckStop)
+			return nil
+		default:
+		}
+
+		e, err := nvml.WaitForEvent(eventSet, 5000)
+		if err != nil && e.Etype != nvml.XidCriticalError {
```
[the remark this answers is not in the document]
ruiwen-zhao: good catch.

pkg/gpu/nvidia/manager.go
```diff
@@ -235,6 +244,106 @@ func (ngm *nvidiaGPUManager) CheckDeviceStates() bool {
...
+			continue
+		}
+
+		if e.Edata == 31 || e.Edata == 43 || e.Edata == 45 {
```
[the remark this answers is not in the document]
ruiwen-zhao: will add https://docs.nvidia.com/deploy/xid-errors/index.html#topic_4

pkg/gpu/nvidia/manager.go
ruiwen-zhao: without this the unit test will fail, because if the health monitoring is not enabled, `nvidiaGPUManager.Stop()` will be blocked indefinitely on `<-ngm.healthCheckStop`. So we need a way for `nvidiaGPUManager.Stop()` to know if it needs to wait for `healthCheckStop` channel to close

pkg/gpu/nvidia/manager.go
ruiwen-zhao: Yeah good catch, even though the behavior will not change (i.e. `ListAndWatch()` will be blocked and never update devices with kubelet), i think it will be better to close it here. 

And even if we close the channel here, I think it is fine to have `ListAndWatch()` call `sendDevice()` when the channel is closed.

pkg/gpu/nvidia/manager.go
```diff
@@ -339,5 +448,18 @@ func (ngm *nvidiaGPUManager) Stop() error {
...
 	ngm.stop <- true
 	<-ngm.stop
+	if ngm.healthCheckRunning {
+		ngm.healthCheckStop <- true
+		<-ngm.healthCheckStop
+	}
 	return nil
 }
+
+func deviceNameFromPath(path string) (string, error) {
```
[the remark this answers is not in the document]
ruiwen-zhao: yeah i will move this to a util package.

ruiwen-zhao:
> Can you help me understand how the health check will work when MIG mode is enabled, and GPUs are partitioned?

Yeah i would like to discuss this with you coz I havent tested the MIG mode yet. But if I am understanding correctly, the NVML functions can all work with MIGs, so we can register events for each MIG instance. The only thing left is that we need to pass the health information to MIG device manager, because now MIG is done in a separate device manager

pkg/gpu/nvidia/health_check/health_checker.go
```diff
@@ -0,0 +1,164 @@
...
+
+	glog.Infof("Found %d GPU devices", count)
+	for i := uint(0); i < count; i++ {
+		device, err := nvml.NewDeviceLite(i)
+		if err != nil {
+			return fmt.Errorf("failed to read device with index %d: %v", i, err)
+		}
+		deviceName, err := util.DeviceNameFromPath(device.Path)
+		if err != nil {
+			glog.Errorf("Invalid GPU device path found: %s. Skipping this device", device.Path)
```
pradvenkat, line 65: Needs a `continue` statement after this line.

pkg/gpu/nvidia/health_check/health_checker.go
```diff
@@ -0,0 +1,164 @@
...
+		}
+
+		if _, ok := hc.devices[deviceName]; !ok {
+			// we only monitor the devices passed in
+			glog.Warningf("Ignoring device %s for health check.", deviceName)
+			continue
+		}
+
+		glog.Infof("Found device %s for health monitoring. UUID: %s", deviceName, device.UUID)
+		hc.nvmlDevices[deviceName] = device
```
pradvenkat, line 75: So my understanding is that when MIG mode is enabled, `hc.nvmlDevices` will be empty, because device name we discover from nvml won't match the MIG device names we use (e.g. `gpu0/gi1`) on line 68. 

So this means when MIG mode is enabled and GPUs are partitioned, there will be no health checks?  That's fine for now, but just want to confirm the behavior.

pkg/gpu/nvidia/health_check/health_checker.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+
+	hc.eventSet = nvml.NewEventSet()
+	for _, d := range hc.nvmlDevices {
+		gpu, _, _, err := nvml.ParseMigDeviceUUID(d.UUID)
```
pradvenkat, line 80: If my above comment (i.e. no health checks for MIG devices), then is this even required? Can we just directly use UUID?

pkg/gpu/nvidia/health_check/health_checker.go
```diff
@@ -0,0 +1,164 @@
...
+	go func() {
+		if err := hc.listenToEvents(); err != nil {
+			glog.Errorf("GPUHealthChecker listenToEvents error: %v", err)
+		}
+	}()
+
+	return nil
+}
+
+// chechHealth registers event with NVML to detect GPU critical errors
```
pradvenkat, line 106: Outdated comment.

pkg/gpu/nvidia/health_check/health_checker.go
```diff
@@ -0,0 +1,164 @@
...
+		select {
+		case <-hc.stop:
+			close(hc.stop)
+			return nil
+		default:
+		}
+
+		e, err := nvml.WaitForEvent(hc.eventSet, 5000)
+		if err != nil || e.Etype != nvml.XidCriticalError {
+			continue
```
pradvenkat, line 118: If `err != nil` should we print the error? maybe with lower verbosity.

pkg/gpu/nvidia/health_check/health_checker.go
```diff
@@ -0,0 +1,164 @@
...
+				hc.health <- d
+			}
+			continue
+		}
+
+		for _, d := range hc.devices {
+			// Please see https://github.com/NVIDIA/gpu-monitoring-tools/blob/148415f505c96052cb3b7fdf443b34ac853139ec/bindings/go/nvml/nvml.h#L1424
+			// for the rationale why gi and ci can be set as such when the UUID is a full GPU UUID and not a MIG device UUID.
+			uuid := hc.nvmlDevices[d.ID].UUID
+			gpu, gi, ci, err := nvml.ParseMigDeviceUUID(uuid)
```
pradvenkat, line 142: Same as earlier, will any of the MIG logic even work?

pkg/gpu/nvidia/health_check/health_checker.go
```diff
@@ -0,0 +1,164 @@
...
+			}
+		}
+	}
+}
+
+// Stop deletes the NVML events and stops the listening go routine
+func (hc *GPUHealthChecker) Stop() {
+	nvml.DeleteEventSet(hc.eventSet)
+	hc.stop <- true
+	<-hc.stop
```
pradvenkat, line 163: If we receive the value here,  would we miss the notification in line 110?

pkg/gpu/nvidia/util/util.go
```diff
@@ -0,0 +1,29 @@
...
+// limitations under the License.
+
+package util
+
+import (
+	"fmt"
+	"regexp"
+)
+
+func DeviceNameFromPath(path string) (string, error) {
```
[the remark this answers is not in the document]
ruiwen-zhao: will do

pkg/gpu/nvidia/health_check/health_checker.go
ruiwen-zhao: oops good catch

pkg/gpu/nvidia/health_check/health_checker.go
ruiwen-zhao: your understanding is correct. If `DeviceNameFromPath(migPath)` does not match the device names passed in, then the mig device will be ignored.

pkg/gpu/nvidia/health_check/health_checker.go
ruiwen-zhao: yeah I left this here mainly for future reference. If we change the device naming pattern for MIG devices, this should work for MIG devices.

pkg/gpu/nvidia/health_check/health_checker.go
ruiwen-zhao: I didnt find the code exactly but i think this error most likely will always be the timeout errors. (WaitForEvent has a 5 second timeout)

pkg/gpu/nvidia/health_check/health_checker.go
ruiwen-zhao: this should work if `DeviceNameFromPath(migPath)` matches the device names passed in. So that's why i left this part here

pkg/gpu/nvidia/health_check/health_checker.go
pradvenkat: Sounds good, can you add a comment or TODO somewhere in this file to tell the reader that MIG devices are not supported yet?

pkg/gpu/nvidia/health_check/health_checker.go
pradvenkat: sounds good, in that case we can leave this as is.

pkg/gpu/nvidia/health_check/health_checker.go
ruiwen-zhao: I dont quite follow. The purpose of `<-hc.stop` is to wait for the stop channel to be closed, which means we are just making sure `listenToEvents()` has caught the signal and will return.

pradvenkat: Thanks for clarifying.

pradvenkat:
Looks great!

pradvenkat (approved):

ruiwen-zhao:
The continuous-integration/travis-ci/pr test cannot run for some reason, but i have run the commands locally and made sure they passed.

```
ruiwen@ruiwen:~/GoogleCloudPlatform/container-engine-accelerators$ make presubmit
git>> vetting code
>> checking go formatting
>> checking file boilerplate
ruiwen@ruiwen:~/GoogleCloudPlatform/container-engine-accelerators$ make test
>> running tests
?   	github.com/GoogleCloudPlatform/container-engine-accelerators/cmd/nvidia_gpu	[no test files]
ok  	github.com/GoogleCloudPlatform/container-engine-accelerators/partition_gpu	(cached)
ok  	github.com/GoogleCloudPlatform/container-engine-accelerators/pkg/gpu/nvidia	(cached)
?   	github.com/GoogleCloudPlatform/container-engine-accelerators/pkg/gpu/nvidia/health_check	[no test files]
?   	github.com/GoogleCloudPlatform/container-engine-accelerators/pkg/gpu/nvidia/metrics	[no test files]
ok  	github.com/GoogleCloudPlatform/container-engine-accelerators/pkg/gpu/nvidia/mig	(cached)
ok  	github.com/GoogleCloudPlatform/container-engine-accelerators/pkg/gpu/nvidia/util	0.061s
```

PR #176, opened 2021-01-22
Add a demo that triggers illegal memory access

ruiwen-zhao:


demo/gpu-error/illegal-memory-access/vectorAdd.cu
```diff
@@ -0,0 +1,201 @@
...
+/**
+ * Vector addition: C = A + B.
+ *
+ * This sample is a very basic sample that implements element by element
+ * vector addition. It is the same as the sample illustrating Chapter 2
+ * of the programming guide with some additions like error checking.
+ */
+/*
+ * hahahah
+ */
```
pradvenkat, line 21: nit: remove

PR #179, opened 2021-01-26
Driver installer daemonset to support MIG

pradvenkat:
This daemonset does two things:

1.  Installs nvidia driver 450.80.02, which is required as per MIG user
guide
(https://docs.nvidia.com/datacenter/tesla/mig-user-guide/index.html#system-considerations).
2.  Invokes the partition_gpu tool to enable MIG mode and create gpu and
compute instances as specified in the GPU configuration file.


pradvenkat:
The build failure is unrelated to this PR, and is being fixed in https://github.com/GoogleCloudPlatform/container-engine-accelerators/pull/175.

PR #182, opened 2021-01-28
Add useful logs for partition-gpu tool

pradvenkat:


pkg/gpu/nvidia/manager.go
```diff
@@ -131,7 +131,7 @@ func (ngm *nvidiaGPUManager) discoverGPUs() error {
 			continue
 		}
 		if reg.MatchString(f.Name()) {
-			glog.Infof("Found Nvidia GPU %q\n", f.Name())
+			glog.V(3).Infof("Found Nvidia GPU %q\n", f.Name())
```
ruiwen-zhao, line 134: why not default verbosity here? it will only be run once right?

pradvenkat: This is actually called from manager.Start(), which can fail due to other reasons (like MIG mode is not set up etc). So I see this keep showing up in the device plugin logs, and it's very noisy.  When the GPUs are discovered, there is a separate log that prints all the discovered devices (including MIG devices), I think that should be sufficient to understand what's happening with the device plugin.

ruiwen-zhao (approved):

PR #193, opened 2021-03-22
Add a demo device plugin yaml with health monitoring enabled based on…

ruiwen-zhao:
… ver 1.0.11

With this demo in place, customers that would like to try out health monitoring can update their daemonset with 

```
kubectl apply -f https://raw.githubusercontent.com/GoogleCloudPlatform/container-engine-accelerators/master/demo/device-plugin-1-0-11-health-monitoring-enabled.yaml
```


pradvenkat:
Do we need the device plugin (1-0-11) version in the file name? Unless there's a good reason to include it, I suggest we drop it from the filename.

pradvenkat (approved):

PR #203, opened 2021-09-01
remove redundant CheckDevicePaths code

grac3gao-zz:
ngm.CheckDevicePaths() is called in 
```
	for {
		err := ngm.CheckDevicePaths()
		if err == nil {
			if err = ngm.Start(); err == nil {
				break
			}
		}
		// Use non-default level to avoid log spam.
		glog.V(3).Infof("nvidiaGPUManager.CheckDevicePaths() failed: %v", err)
		time.Sleep(5 * time.Second)
	}
```
there is no need to call it again inside of ngm.Start()


pradvenkat (changes requested):
Abandon this? Since this is already addressed in pull/208?

PR #213, opened 2022-01-20
Fix `/home/kubernetes/bin/nvidia is not a directory`

c202c:
If nothing exists at  /home/kubernetes/bin/nvidia, an empty directory will be created.

When creating a new GKE cluster with GPU, the pods of `nvidia-gpu-device-plugin` DaemonSet keep `ContainerCreating` status. With the following messages:
```
MountVolume.SetUp failed for volume "nvidia" : hostPath type check failed: /home/kubernetes/bin/nvidia is not a directory
```


zsolt-p:
I'm facing this same problem and this looks like a valid fix, thanks @c202c 
Can't find a related issue on the tracker, how can I support moving this ahead?

PR #214, opened 2022-01-21
Add MIG/MIG+time-sharing UT for plugin

grac3gao:
1.) Add MIG/MIG+time-sharing UT for plugin
2.) Refactor existing UTs

Noted:
1). alpha_plugin doesn't include any MIG UTs, as MIG is not supported in alpha plugin

### DIFF
--- a/cmd/nvidia_gpu/nvidia_gpu.go
+++ b/cmd/nvidia_gpu/nvidia_gpu.go
@@ -33,6 +33,8 @@ const (
 	kubeletEndpoint      = "kubelet.sock"
 	pluginEndpointPrefix = "nvidiaGPU"
 	devDirectory         = "/dev"
+	// Proc directory is used to lookup the access files for each GPU partition.
+	procDirectory = "/proc"
 )
 
 var (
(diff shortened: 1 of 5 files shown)


pkg/gpu/nvidia/manager.go
```diff
@@ -83,7 +81,7 @@ type MountPath struct {
 	ContainerPath string
 }
 
-func NewNvidiaGPUManager(devDirectory string, mountPaths []MountPath, gpuConfig GPUConfig) *nvidiaGPUManager {
+func NewNvidiaGPUManager(devDirectory, procDirectory string, mountPaths []MountPath, gpuConfig GPUConfig) *nvidiaGPUManager {
```
grac3gao, line 84: I changed procDirectory to be an input when initializing NvidiaGPUManager, so that the unit test can directly replace this procDirectory with a test path. Changing the procDirectory to be an input doesn't change the existing logic, and it help to save effort to write a set of mock functions for overriding (otherwise, the procDirectory is unchangeable).

cmd/nvidia_gpu/nvidia_gpu.go
```diff
@@ -87,7 +89,7 @@ func main() {
 		}
 	}
 	glog.Infof("Using gpu config: %v", gpuConfig)
-	ngm := gpumanager.NewNvidiaGPUManager(devDirectory, mountPaths, gpuConfig)
+	ngm := gpumanager.NewNvidiaGPUManager(devDirectory, procDirectory, mountPaths, gpuConfig)
```
grac3gao, line 92: I changed procDirectory to be an input when initializing NvidiaGPUManager, so that the unit test can directly replace this procDirectory with a test path. Changing the procDirectory to be an input doesn't change the existing logic, and it helps to save effort to write a set of mock functions for overriding (otherwise, the procDirectory is unchangeable).

pradvenkat (approved):

PR #216, opened 2022-01-21
Add NVIDIA MPS support to device plugin

pradvenkat:


pkg/gpu/nvidia/manager.go
```diff
@@ -58,13 +69,58 @@ var (
...
+	} else {
+		switch config.GPUSharingConfig.GPUSharingStrategy {
+		case TimeSharing, MPS:
+			if config.GPUSharingConfig.MaxSharedClientsPerGPU <= 0 {
+				return fmt.Errorf("MaxSharedClientsPerGPU should be > 0 for time-sharing or mps GPU sharing strategies")
+			}
+			break
+		case Undefined:
+			if config.GPUSharingConfig.MaxSharedClientsPerGPU > 0 {
+				return fmt.Errorf("GPU sharing strategy needs to specified when MaxSharedClientsPerGPU > 0")
```
grac3gao, line 110: nit: needs to specified -> needs to be specified

pkg/gpu/nvidia/manager.go
```diff
@@ -112,10 +164,10 @@ func (ngm *nvidiaGPUManager) ListDevices() map[string]pluginapi.Device {
 
 	switch {
 	// We will have MPS solution in the future, which will be mutually exclusive with the time-sharing solution.
-	case ngm.gpuConfig.MaxTimeSharedClientsPerGPU > 0:
+	case ngm.gpuConfig.GPUSharingConfig.MaxSharedClientsPerGPU > 0:
```
grac3gao, line 167: Does this change only mean to change the time-sharing to use the new parameter, or this change mean to expand the usage to time-sharing + MPS?

If it is the former one, it missed to check the sharing strategy is time-sharing.
If it is the latter one:
1). how will the virtual device ID from MPS be in the same form as time-sharing (we need to configure related code for the MPS specific initContainer in driver installer?)
2). we'll need to update the comments reflect it includes MPS, and update the package name `timesharing.VirtualToPhysicalDeviceID` in follow-up PRs.

pkg/gpu/nvidia/beta_plugin.go
```diff
@@ -82,13 +82,11 @@ func (s *pluginServiceV1Beta1) Allocate(ctx context.Context, requests *pluginapi
 			})
 		}
 
-		for _, mountPath := range s.ngm.mountPaths {
-			resp.Mounts = append(resp.Mounts, &pluginapi.Mount{
-				HostPath:      mountPath.HostPath,
-				ContainerPath: mountPath.ContainerPath,
-				ReadOnly:      true,
-			})
+		for i := range s.ngm.mountPaths {
```
grac3gao, line 85: line 60 in this file, the package name is `timesharing.ValidateRequest`, but it applies to both MPS and timesharing (if the virtual ID naming rule is the same), we'll need to update the package name as a follow-up

pkg/gpu/nvidia/manager.go
pradvenkat: It's the latter.  Basically MPS is a process running on the host, that has exclusive control of the GPUs. However MPS is transparent to any CUDA application (i.e. application doesn't need to know whether MPS is running or not). The only thing necessary for containers to consume the GPU through MPS is the directory used by MPS for named pipes and unix sockets. This additional directory is also mounted by the device plugin, but the rest of the device discovery and consumption remains unchanged.

Agree that we need to change the timesharing package name, will update that.

pkg/gpu/nvidia/beta_plugin.go
pradvenkat: Forgot about it, thanks for pointing it :).

grac3gao (approved):

PR #219, opened 2022-02-03
Delete alpha plugin from pkg/gpu/nvidia

richardsliu:


pkg/gpu/nvidia/multiple_versions_test.go
```diff
@@ -134,16 +132,4 @@ func TestNvidiaGPUManagerMultuipleAPIs(t *testing.T) {
 			{DevicesIDs: []string{"nvidia1", "nvidia3"}}}})
 	as.Nil(resp)
 	as.NotNil(err)
-
-	// Tests Alpha ListAndWatch
-	stream2, err := clientAlpha.ListAndWatch(context.Background(), &pluginalpha.Empty{})
```
pradvenkat, line 139: Do we still need multiple_versions_test? Once alpha is deprecated, beta API is the only supported one.  Can you confirm if this test is already covered by beta_plugin_test? If yes, then maybe we can remove this too?

richardsliu: I think this test only covers ListAndWatch which is already covered by beta_plugin_test. But should we leave it here anyway in case we are planning to add more API versions later?

pradvenkat: My suggestion is if it doesn't add value now, let's delete it. We can add it back later if needed, but it's unlikely we're going to have a new version any time soon :)

richardsliu: @pradvenkat Done.

pradvenkat (approved):

PR #226, opened 2022-03-16
add fastsocket_installer code

melody789:
Add fastsocket installer image builder

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,6 +38,7 @@ TAG=$(shell cat VERSION)
 REGISTRY?=gcr.io/google-containers
 IMAGE=nvidia-gpu-device-plugin
 PARTITION_GPU_IMAGE=nvidia-partition-gpu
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -51,4 +52,7 @@ push:
 partition-gpu:
 	docker build --pull -t ${REGISTRY}/${PARTITION_GPU_IMAGE}:${TAG} -f partition_gpu/Dockerfile .
 
+fastsocket_installer:
+	docker build --pull -t ${REGISTRY}/${FASTSOCKET_INSTALLER_IMAGE}:${TAG} -f fast-socket-installer/Dockerfile .
```
richardsliu, line 56: Should we use a different tag here?

melody789: I put the tag here for testing purpose. I would like to see how much difference from 'partition-gpu'. I will change the tag later after the louhi flow works.

PR #241, opened 2022-07-12
Add new mig partition for the new accelerator type

crystalzhaizhai:
A100-SXM4-80GB


fmoessbauer:
@crystalzhaizhai May I ask you to have a look at #242 . The linked PR provides a generic solution which should work for all MIG capable cards.

PR #247, opened 2022-09-07
Change NCCL fast socket base image

melody789:
Change NCCL fast socket base image

### DIFF
--- a/fast-socket-installer/image/Dockerfile
+++ b/fast-socket-installer/image/Dockerfile
@@ -1,7 +1,7 @@
-FROM k8s.gcr.io/debian-base:v1.0.0
+FROM gcr.io/gke-release/debian-base:latest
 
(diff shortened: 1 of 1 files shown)


fast-socket-installer/image/Dockerfile
```diff
@@ -1,7 +1,7 @@
-FROM k8s.gcr.io/debian-base:v1.0.0
+FROM gcr.io/gke-release/debian-base:latest
```
richardsliu, line 1: Use a specific digest or tag instead of latest, otherwise things could be broken unexpectedly.

melody789: I see. I changed to the newest tag.

melody789: I see. I just changed it to the newest tag.

PR #255, opened 2022-10-20
Update metrics to avoid possible segfaults

grac3gao:
Update metrics to avoid potential segfaults when gmc.collectStatus return error


elfinhe (approved):

elfinhe (approved):
/LGTM

PR #259, opened 2022-11-17
Combine the installer for normal GPU and MIG GPU

grac3gao:
Combine the installer for normal GPU and MIG GPU. The additional init container for MIG (partition-gpus) will be skipped when the installer is running in the normal GPU. There is no increasing installer latency detected for normal GPU with this change.


melody789 (approved):
LGTM

PR #262, opened 2022-12-05
xid error mitigation

crystalzhaizhai:
This is making xid error mitigation public

### DIFF
--- a/demo/xid-error-mitigation.yaml
+++ b/demo/xid-error-mitigation.yaml
@@ -1,0 +1,140 @@
+## Opt-in Error Codes Detection in Device Plugin Health Checker
+
+### Why do we want to make it opt-in rather than hard-coded in device plugin?
+
+If it's a hardware issue, we want to catch that in the health check and kubectl
+describe nodes will show it to customers so that they can migrate their
+workloads. The next step is GKE migrate the workload through graceful node
+shutdown.
+
+xid 79 could be a software issue. Migrating the workload to other nodes could
+have the same issue. We only want health check code to catch hardware errors, so
+we won't hardcode 79 in the health check code. We want the customers to do their
+investigation to rule out the possible user errors before treating it as a
+hardware failure. Once the customers actively opt in with configuration files,
+we assume they are aware of the possible user errors and they intend to treat 79
+as a hardware failure.
+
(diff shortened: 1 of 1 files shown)


demo/xid-error-mitigation.yaml
```diff
@@ -0,0 +1,140 @@
...
+as a hardware failure.
+
+### Xid Code ConfigMap
+
+Create a yaml file like this and put the xid codes you intend to add under
+data.HealthCriticalXid with the following format. Apply the yaml file to your
+cluster.
+
+```
+apiVersion: v1
```
richardsliu, line 25: Can we add these as actual files in a directory with a README.md file explaining how to deploy it? Then the user could just clone the repo and deploy it from their commandline.

grac3gao (commented):
This is a short-term mitigation which can't be used in a long run. The device plugin change introduced in this doc will be reverted by the add-on manager during an upgrade. If the GPU nodes are used together with autoscaling, new nodes may not contain this mitigation.

Ideally, we would like to have a more stable and simpler solution for this problem. (e.g. user only need to configure the configmap to customize the XID they like).

It would be better to add more explanation in this doc for this situation, mentioning it is a short-term mitigation.

thomas-riccardi:
> Ideally, we would like to have a more stable and simpler solution for this problem. (e.g. user only need to configure the configmap to customize the XID they like).

We would like that, yes: just edit/create a configmap.

For now, we can modify the `/etc/nvidia/gpu_config.json` file on the host, but we have no guarantee it's done *before* the gpu-device-plugin loads it, and have no easy way to reload it if it's done after. (alternatively, gpu-device-plugin could watch it?)

Also, xid 79 should IMO be there by default: it's *not* a user error: https://docs.nvidia.com/deploy/xid-errors/index.html#topic_4

Context: we get `NVRM: Xid (PCI:0000:00:04): 79, pid=0, GPU has fallen off the bus.` on GKE, and the node stay stuck with a unusable GPU, our workload using the GPU at best reaches a CRITICAL/FATAL state and crash-loops, or at worse gets silently stuck. 
The end goal is that this would somehow mark the node as broken, so the GKE auto-repair feature kicks-in.

PR #263, opened 2022-12-11
add option to mount nvidia tools into container

fmoessbauer:
This patch adds support to mount the nvidia-tools like nvidia-smi into the container. A new parameter host-tool-path is added to control where the nvidia tools are located on the host.

Only whitelisted tools are mounted into the container and that is done on a best-effort basis, as the tools might be dynamically linked.

Signed-off-by: Felix Moessbauer <felix.moessbauer@siemens.com>


fmoessbauer:
Any news on this?

Issue #265, opened 2022-12-12
unknown field "spec.template.spec.initContainers[0].volumeMounts[5].hostPath


maxpain:
Hello.

This daemonset  is invalid:
https://github.com/GoogleCloudPlatform/container-engine-accelerators/blob/master/nvidia-driver-installer/cos/daemonset-preloaded-latest.yaml#L106

```bash
maxpain@Maksims-MacBook-Pro-4 ~ % kubectl apply -f https://raw.githubusercontent.com/GoogleCloudPlatform/container-engine-accelerators/master/nvidia-driver-installer/cos/daemonset-preloaded-latest.yaml
Error from server (BadRequest): error when creating "https://raw.githubusercontent.com/GoogleCloudPlatform/container-engine-accelerators/master/nvidia-driver-installer/cos/daemonset-preloaded-latest.yaml": DaemonSet in version "v1" cannot be handled as a DaemonSet: strict decoding error: unknown field "spec.template.spec.initContainers[0].volumeMounts[5].hostPath"
```

capfish:
It looks like this was just fixed in https://github.com/GoogleCloudPlatform/container-engine-accelerators/commit/3e03730c01253feb63af1feafcbc413f0eba5742

PR #266, opened 2022-12-13
#265 moved misplaced volume spec to within volumes

capfish:
It looks like someone just put a volume spec in the wrong place.  This matches the yaml configuration of the not "-latest" version which does work as expected.


capfish:
Looks like this was just fixed with
https://github.com/GoogleCloudPlatform/container-engine-accelerators/commit/3e03730c01253feb63af1feafcbc413f0eba5742

PR #271, opened 2022-12-14
Change NCCL daemonset name

melody789:
Change the NCCL DaemonSet name for GKE 1.24 version

### DIFF
--- a/fast-socket-installer/fast-socket-installer.yaml
+++ b/fast-socket-installer/fast-socket-installer.yaml
@@ -1,21 +1,21 @@
 apiVersion: apps/v1
 kind: DaemonSet
 metadata:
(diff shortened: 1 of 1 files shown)


fast-socket-installer/fast-socket-installer.yaml
```diff
@@ -1,21 +1,21 @@
 apiVersion: apps/v1
 kind: DaemonSet
 metadata:
-  name: nccl-fastsocket-installer
+  name: nccl-fastsocket-installer2
```
richardsliu, line 4: Can we use a better name for this? Like 'nccl-fastsocket-manual-installer' or something.

melody789: Yes we can

PR #285, opened 2023-03-04
Migrate podresources API from v1alpha to v1

danielvegamyhre:
Migrate podresources API from v1alpha to v1


danielvegamyhre:
Closing PR for now since v1 won't be available until 1.27

PR #291, opened 2023-04-26
Use go-nvml library for metrics

grac3gao:
Only updated the new library for metrics.
Health check keeps the old library.

Tested with new GPU, the metrics library works.
Tested with A100, the new library exported metrics has the same value as the old library.
Tested with A100 and induced XID 31, the health check works fine.


elfinhe (approved):

pkg/gpu/nvidia/metrics/metrics.go
```diff
@@ -175,53 +171,58 @@ func (m *MetricServer) collectMetrics() {
 	}
 }
 
-func getGpuMetrics(device string, d *nvml.Device) (uint, uint64, error) {
-	status, err := gmc.collectStatus(d)
-	if err != nil {
-		glog.Errorf("Failed to get device status for %s: %v", device, err)
-		return 0, 0, err
+// getGpuMetricsInfo returns dutyCycle, usedMemory, totalMemory, uuid and deviceModel.
+func getGpuMetricsInfo(device string, d *nvml.Device) (uint, uint64, uint64, string, string, error) {
```
samuelkarp, line 175: nit: there are enough return values for this function that a struct might be help the calling code be more clear

pkg/gpu/nvidia/metrics/metrics.go
```diff
@@ -175,53 +171,58 @@ func (m *MetricServer) collectMetrics() {
...
+	if ret != nvml.SUCCESS {
+		return 0, 0, 0, "", "", fmt.Errorf("failed to get GPU memory: %v", nvml.ErrorString(ret))
+	}
+	dutyCycle, err := gmc.collectDutyCycle(uuid, time.Second*10)
 	if err != nil {
-		return 0, 0, fmt.Errorf("Failed to get dutyCycle: %v", err)
+		return 0, 0, 0, "", "", fmt.Errorf("failed to get dutyCycle: %v", err)
 	}
-	return dutyCycle, *mem.Global.Used, nil
+	return dutyCycle, mem.Used / (1024 * 1024), mem.Total / (1024 * 1024), uuid, deviceModel, nil
```
samuelkarp, line 193: nit: a const for `1024 * 1024` to make the unit conversion more clear can help

something like:

```go
const mib = 1024 * 1024
return dutyCycle, mem.Used / mib, mem.Total / mib, uuid, deviceModel, nil
```

samuelkarp (commented):
LGTM, a couple nits around clarity but no urgency in addressing those

PR #295, opened 2023-05-09
Update to latest device plugin manifest

grac3gao:
Update to latest device plugin manifest


elfinhe (commented):
is it possible we keep previous format?

The diff of the PR is not easy to figure out what has been changed.

elfinhe (approved):

PR #316, opened 2023-08-15
Update GPU driver isntaller daemonset to check for existing driver mo…

Jiaqicao257:
…dules


grac3gao (dismissed):

samuelkarp (approved):

elfinhe (approved):

arnav-kansal (approved):
This check should work!

This DS will exit silently and user may expect the driver version to change.

Can we add some logs to say there is an existing GPU driver version that is installed that may or may not match the driver being tried to install.

grac3gao (approved):

PR #322, opened 2023-10-05
Remove check for existing driver modules in driver installer

Jiaqicao257:


elfinhe (approved):

grac3gao:
Do we want to wait until this logic is configured inside the installer image, then remove it from the command line?

Jiaqicao257:
> Do we want to wait until this logic is configured inside the installer image, then remove it from the command line?

Sounds good

PR #337, opened 2023-11-30
Update nccl-test manifest

grac3gao:
Update nccl-test manifest to run nccl-test with a simpler workaround


gpudirect-tcpx/nccl-test.yaml
```diff
@@ -1,143 +1,90 @@
-apiVersion: v1
-kind: Service
```
aojea, line 2: you no longer need the services?

grac3gao: We still need the node name to run the nccl-test. The manifest used to configure the service, because it needs to ssh into the node to run some commands. And running a fixed service name is simpler than checking the node name every time and copy paste into the command. 

For now, there are some simpler workaround, which are updated in the user guide.

gpudirect-tcpx/nccl-test.yaml
```diff
@@ -1,143 +1,89 @@
...
+              service ssh restart;
+              sleep infinity;
+          env:
+            - name: LD_LIBRARY_PATH
+              value: /usr/local/nvidia/lib64
+          securityContext:
+            privileged: true
+            capabilities:
+              add:
+                - IPC_LOCK
```
samuelkarp, line 74: I don't think these are needed anymore?

grac3gao: Yeah, the `privileged:true` could be removed from the nccl-test container. It only needs IPC_LOCK capabilities. Removed. Thanks for catching this!

samuelkarp: `IPC_LOCK` should also be able to be removed.

grac3gao: Thanks! Validated and removed.

samuelkarp (approved):

PR #343, opened 2023-12-20
Update Makefile

luisvillarreal:
Added ubuntu 22.04 for newer versions


CoderSherlock:
@luisvillarreal Could you please update CLA ?

PR #345, opened 2023-12-21
Add NRI device injector package

Jiaqicao257:


nri_device_injector/nri_device_injector.go
```diff
@@ -0,0 +1,191 @@
...
+		ctrName string
+		devices []device
+		err     error
+	)
+
+	ctrName = container.Name
+	log.Infof("Started CreateContainer on container: %v, pod: %v, namespace: %v", ctrName, pod.Name, pod.Namespace)
+	if pod != nil {
+		ctrName = pod.Name + "/" + ctrName
+	}
```
aojea, line 93: this logic deserves a comment, based on the tests it seems that you can pass either pod level or container level devices?

nri_device_injector/nri_device_injector.go
```diff
@@ -0,0 +1,191 @@
...
+			if err != nil {
+				return nil, nil, err
+			}
+			adjust.AddDevice(deviceNRI)
+			log.Infof("Container %s: injected device %q...", ctrName, d.Path)
+		}
+	}
+
+	return adjust, nil, nil
+}
```
aojea, line 117: nit for readability usually golang avoid multiple nesting
```suggestion
	devices, err = getDevices(container.Name, pod.Annotations)
	if err != nil {
		return nil, nil, err
	}
	adjust := &api.ContainerAdjustment{}

	if len(devices) == 0 {
		log.Infof("Container %s: no devices annotated...", ctrName)
		return adjust, nil, nil
	}
        for _, d := range devices {
	log.Infof("Container %s: annotated device %q...", ctrName, d.Path)
	deviceNRI, err := d.toNRIDevice()
	if err != nil {
		return nil, nil, err
	}
	adjust.AddDevice(deviceNRI)
	log.Infof("Container %s: injected device %q...", ctrName, d.Path)

	return adjust, nil, nil
}
```

nri_device_injector/nri_device_injector.go
```diff
@@ -0,0 +1,191 @@
...
+	}
+
+	if len(devices) == 0 {
+		log.Infof("Container %s: no devices annotated...", ctrName)
+	} else {
+		for _, d := range devices {
+			log.Infof("Container %s: annotated device %q...", ctrName, d.Path)
+			deviceNRI, err := d.toNRIDevice()
+			if err != nil {
+				return nil, nil, err
```
aojea, line 109: is there any difference on returning nil or the empty value here?

nri_device_injector/nri_device_injector_test.go
```diff
@@ -0,0 +1,104 @@
...
+		"One valid device annotation injecting to container": {
+			container: "test",
+			annotations: map[string]string{
+				"devices.gke.io/container.test": `
+- path: /dev/test
+`}, want: []device{{
+				Path: "/dev/test",
+			}},
+		},
+		"One valid device annotation injecting to pod": {
```
aojea, line 46: @samuelkarp we had this conversation in the past, I think that for network devices this model will be more simple than existing one, IIUIC the only concern was that the OCI spec does not consider this devices? do you think that is a debate we can open?

nri_device_injector/nri_device_injector.go
```diff
@@ -0,0 +1,191 @@
...
+	for _, key = range []string{
+		deviceKeyPrefix + "/container." + ctrName,
+		deviceKeyPrefix + "/pod",
+		deviceKeyPrefix,
+	} {
+		if value, ok := podAnnotations[key]; ok {
+			annotation = []byte(value)
+			break
+		}
+	}
```
aojea, line 136: can there be multiple annotations and we only pick them by this order of preference, first container, second pod , ... ? 
if that is the logic can we add a comment explaining it?

nri_device_injector/nri_device_injector.go
Jiaqicao257: I added the comment on top of the file where the device injection keys were defined, and will follow up with a readme under the directory to include more detailed instructions.

nri_device_injector/nri_device_injector.go
Jiaqicao257: Done, thanks!

nri_device_injector/nri_device_injector.go
Jiaqicao257: There can be multiple annotations, and here I'm making container level and pod level specs exclusive to each other to reduce complexity. I added some comments to explain this.

samuelkarp: Do we have a need for pod-level injection?  If we're really only looking at container-level for our immediate use-cases, it'd be simpler to remove the pod-level logic for now and add it back if we ever have a future need.

nri_device_injector/nri_device_injector.go
Jiaqicao257: I don't think so, as long as the plugin returns error, there won't be any adjustment made.

nri_device_injector/nri_device_injector_test.go
samuelkarp: > IIUIC the only concern was that the OCI spec does not consider netdevices? do you think that is a debate we can open?

I think that'd be a reasonable discussion to have.  https://github.com/opencontainers/runtime-spec is the right place to open an issue.  https://github.com/opencontainers/runtime-spec/blob/main/config-linux.md describes what is supported both from a namespace (including network namespace) perspective and devices.

nri_device_injector/nri_device_injector.go
Jiaqicao257: @samuelkarp Makes sense, I've removed the pod level injection and only allow container level injection. Thanks!

nri_device_injector/nri_device_injector.go
```diff
@@ -0,0 +1,184 @@
...
+// The plugin makes adjustment on containers with device injection annotations.
+func (p *plugin) CreateContainer(_ context.Context, pod *api.PodSandbox, container *api.Container) (*api.ContainerAdjustment, []*api.ContainerUpdate, error) {
+	var (
+		ctrName string
+		devices []device
+		err     error
+	)
+
+	ctrName = container.Name
+	log.Infof("Started CreateContainer on container: %v, pod: %v, namespace: %v", ctrName, pod.Name, pod.Namespace)
```
samuelkarp, line 93:
```suggestion
	l := log.WithField("container", ctrName).WithField("pod", pod.Name).WithField("namespace", pod.Namespace)
	l.Info("Started CreateContainer")
```

We can then reuse `l` to keep this context throughout the rest of the function.

(If we end up wanting to pass it around elsewhere, the pattern used inside containerd is `github.com/containerd/log` which associates a logger with a `context.Context` and then accesses with `log.G(ctx)...`.)

nri_device_injector/nri_device_injector.go
```diff
@@ -0,0 +1,184 @@
...
+	if pod != nil {
+		ctrName = pod.Name + "/" + ctrName
+	}
```
samuelkarp, line 96: if pod == nil, perhaps we want to return and skip the rest of the logic?  We continue to dereference it on line 98, so a nil pod would cause a panic.

```go
if pod == nil {
	return nil, nil, nil
}
ctrName = pod.Name + "/" + ctrName
```

nri_device_injector/nri_device_injector.go
```diff
@@ -0,0 +1,184 @@
...
+
+	devices, err = getDevices(container.Name, pod.Annotations)
+	if err != nil {
+		return nil, nil, err
+	}
+	adjust := &api.ContainerAdjustment{}
+
+	if len(devices) == 0 {
+		log.Infof("Container %s: no devices annotated...", ctrName)
```
samuelkarp, line 105: nit: we probably want this at debug level rather than info

nri_device_injector/README.md
Jiaqicao257, line 1: @samuelkarp @aojea added readme for the directory, PTAL, thanks!

nri_device_injector/nri_device_injector.go
```diff
@@ -0,0 +1,184 @@
...
+func (d *device) toNRIDevice() (*api.LinuxDevice, error) {
+	var (
+		stat unix.Stat_t
+	)
+	if err := unix.Lstat(d.Path, &stat); err != nil {
+		return nil, err
+	}
+
+	var (
+		devNumber = uint64(stat.Rdev) //nolint:nolintlint,unconvert // the type is 32bit on mips.
```
samuelkarp, line 152: I don't think we're running a linter here? unconvert or nolintlint?

(Adding a linter is a good idea, but if we don't have one let's remove the directive for now and add it back when we need it.)

nri_device_injector/nri_device_injector_test.go
```diff
@@ -0,0 +1,91 @@
...
+
+package main
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestParseDevices(t *testing.T) {
```
samuelkarp, line 23: Not a blocker, but something else we could do: set up some test device nodes with known attributes, and validate `device.toNRIDevice` as well.

nri_device_injector/nri_device_injector_test.go
```diff
@@ -0,0 +1,91 @@
...
+			}, {
+				Path:  "/dev/test1",
+				Major: 123,
+				Minor: 456,
+			}, {
+				Path: "/dev/test2",
+				Type: "b",
+			}},
+		},
+		"Invalid device annotation": {
```
samuelkarp, line 67: Can we add a couple more test cases:
* empty annotation value
* invalid yaml
* devices defined multiple times

samuelkarp (commented):
This looks really good.  A few suggestions, but otherwise looks good.

nri_device_injector/nri_device_injector.go
aojea: great

nri_device_injector/nri_device_injector.go
aojea: it seems this repo hs a linter https://github.com/GoogleCloudPlatform/container-engine-accelerators/actions/runs/7295051489/job/19880874295?pr=345

samuelkarp: Yes, https://github.com/GoogleCloudPlatform/container-engine-accelerators/blob/2ba1c84deee6cb50ede81e132ed10b9ac6d7a748/Makefile#L27-L35, but that wouldn't include the unconvert or nolintlint linters.

nri_device_injector/nri_device_injector.go
Jiaqicao257: Done, thanks!

nri_device_injector/nri_device_injector.go
Jiaqicao257: Done.

nri_device_injector/nri_device_injector.go
Jiaqicao257: Done.

nri_device_injector/nri_device_injector.go
Jiaqicao257: Thanks Sam, I've removed the nolint comment

nri_device_injector/nri_device_injector_test.go
Jiaqicao257: Added. For the devices defined multiple times case, though I added a test case, I didn't really make the logic of getDevice to make sure there's only one annotation for each device path. And when a pod annotates a device multiple time, it's seen as there are multiple nri plugins trying to claim the same device and then the pod gets error creating container (https://github.com/containerd/nri/blob/main/pkg/adaptation/result.go#L1030). Or do we actually want to make sure each device's annotated at most once?

nri_device_injector/nri_device_injector_test.go
Jiaqicao257: By `set up some test device nodes with known attributes`, do you mean manual test on GKE nodes or some tests like https://github.com/opencontainers/runc/blob/main/libcontainer/devices/device_unix_test.go?

samuelkarp: You can create device node files (at any path) using mknod(2) or [`unix.Mknod`](https://pkg.go.dev/golang.org/x/sys/unix#Mknod).  Then you should be able to read them with `unix.Lstat` inside `device.toNRIDevice`.

(mknod(2) requires `CAP_MKNOD`, so that may mean the easiest way to run the test is with `sudo`.  My usual preferred pattern for something like this is to build a test binary using `go test -c` and then run the binary with `sudo`, rather than `sudo go test` that also runs the Go toolchain with privilege.  There's an example of that [here](https://github.com/samuelkarp/runj/blob/4c1dfb645be03f0deb1faae5a551ce6039368d44/Makefile#L31-L39).  It can also be useful to specifically mark a test as requiring root with something like [containerd's `testutil.RequiresRoot`](https://github.com/containerd/containerd/blob/ced9b18c231a28990617bc0a4b8ce2e81ee2ffa1/pkg/testutil/helpers_unix.go#L29-L36) and a [`-test.root` command line flag](https://github.com/containerd/containerd/blob/ced9b18c231a28990617bc0a4b8ce2e81ee2ffa1/Makefile#L214).)

Probably best to do this all in a follow-up PR so you don't block submission on this.

nri_device_injector/nri_device_injector.go
```diff
@@ -0,0 +1,186 @@
...
+		return nil, nil, err
+	}
+	adjust := &api.ContainerAdjustment{}
+
+	if len(devices) == 0 {
+		l.Debug("No devices annotated...")
+		return adjust, nil, nil
+	}
+	for _, d := range devices {
+		l.Infof("Annotated device %q...", d.Path)
```
samuelkarp, line 110: (nit, multiple occurrences)

```suggestion
		l.WithField("device", d.Path).Info("Annotated device")
```

nri_device_injector/nri_device_injector_test.go
samuelkarp: > Or do we actually want to make sure each device's annotated at most once?

Yeah, I think it'd be better to have the plugin return an error if the same device is defined multiple times.

I'm fine with that being done in a follow-up PR rather than this one, so don't consider it a blocker for merging.

samuelkarp (dismissed):

nri_device_injector/nri_device_injector_test.go
aojea: opened https://github.com/opencontainers/runtime-spec/issues/1239

nri_device_injector/nri_device_injector.go
Jiaqicao257: Done.

grac3gao (approved):

nri_device_injector/nri_device_injector_test.go
Jiaqicao257: I see, thanks Sam, I'll create another PR for this.

nri_device_injector/nri_device_injector_test.go
Jiaqicao257: Thanks Sam, I'll create another PR for this.

klueska:
Can someone add some context on what this newly integrated NRI plugin is designed to accomplish? On the surface it seems like CDI is a better fit (and how we now do things in NVIDIA's GPU device plugin).

If we have any hope of eventually merging the GKE GPU plugin with NVIDIA's we will need to understand what is being added here and why it is necessary / how it differs from just using CDI.

klueska:
Also worth noting -- CDI devices were added to the device plugin API in Kubernetes 1.28, graduated to beta in 1.29, and will be promoted to GA in 1.30. And both containerd 1.7+ and cri-o support injecting them natively (without the need for a custom NRI plugin).

Jiaqicao257:
@klueska Hi Kevin, thanks for your comments! This NRI plugin is not related to GPU device plugin itself, but is related to the GPU feature TCPX, that's only supported on GKE 1.27 for now, and this plugin is used to provide unprivileged GPU access on a TCPX sidecar container.

PR #348, opened 2024-01-11
NRI device injector skips duplicate devices 

Jiaqicao257:


nri_device_injector/nri_device_injector.go
```diff
@@ -138,6 +138,14 @@ func getDevices(ctrName string, podAnnotations map[string]string) ([]device, err
 	if err := yaml.Unmarshal(annotation, &devices); err != nil {
 		return nil, fmt.Errorf("invalid device annotation %q: %w", deviceKey, err)
 	}
+	paths := make(map[string]bool)
+	for _, d := range devices {
+		if _, got := paths[d.Path]; got {
+			return nil, fmt.Errorf("duplicate annotation of device with path: %q", d.Path)
```
grac3gao, line 144: Do we want error out (stop the injection process), or just silently skip the duplicated path? Will the user pod show any error if we error out here (injection failed)?

It seems we only use path (even specify minor or major in the annotation) to get the true NTI device info by using unix.Lstat(d.Path, &stat), so if we have two annotation with the same path, it doesn't matter which one we choose.

Jiaqicao257: Make sense, with the previous implementation, container creation will fail if there're several same device injections to the same container. But I agree that we can instead skip the duplicated devices and leave only the first one. I've updated the PR.

grac3gao (approved):

PR #358, opened 2024-03-14
Update nccl test example

grac3gao:
Update nccl test example


Jiaqicao257 (commented):
https://github.com/GoogleCloudPlatform/container-engine-accelerators/blob/master/gpudirect-tcpx/nccl-test-unprivileged-without-hostnetwork.yaml needs to be updated with the same change as well right?

Jiaqicao257 (approved):

PR #369, opened 2024-05-01
Add mknod tests for NRI device injector

Jiaqicao257:


Makefile
```diff
@@ -60,3 +60,14 @@ nri-device-injector:
 	docker build --pull -t ${REGISTRY}/${DEVICE_INJECTOR_IMAGE}:${TAG} -f nri_device_injector/Dockerfile .
 
 .PHONY: all format test vet presubmit build container push partition-gpu
+
+bin/device-injector-test:
+	$(GO) test -c ./nri_device_injector -o ./bin/device-injector-test 
+
+.PHONY: device-injector-test
+device-injector-test: bin/device-injector-test
```
samuelkarp, line 68: Can we run this as part of the GitHub Actions workflow as well?

nri_device_injector/nri_device_injector_test.go
```diff
@@ -15,12 +15,81 @@
...
+	"os"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	"golang.org/x/sys/unix"
 )
 
-func TestParseDevices(t *testing.T) {
+func TestToNRIDevice(t *testing.T) {
+	tempDir, err := os.MkdirTemp("", "device-test")
```
samuelkarp, line 26: You may want to skip this except when explicitly running root-requiring tests.  The pattern we use for this in containerd is https://github.com/containerd/containerd/blob/d9a58a892b77f292b842b849f059d8d8e8972b4a/pkg/testutil/helpers_unix.go#L29-L36

.github/workflows/go_tests.yml
```diff
@@ -22,3 +22,7 @@ jobs:
       run: make presubmit
     - name: Run go test
       run: make test
+    - name: Run device-injector-test
+      run: |
+        echo "${{ secrets.SUDO_PASSWORD }}" | sudo -S make device-injector-test
```
samuelkarp, line 27: Are we using custom actions runners?  It looks like the default GitHub Actions runners have [passwordless sudo enabled](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#administrative-privileges).

nri_device_injector/nri_device_injector_test.go
Jiaqicao257: I see, I'll try remove using the secret then.

Makefile
Jiaqicao257: Done

nri_device_injector/nri_device_injector_test.go
Jiaqicao257: Done, the test is skipped if uid isn't 0.

.github/workflows/go_tests.yml
Jiaqicao257: You're right, tried sudo without the secret directly and it works. Thanks Sam!

nri_device_injector/nri_device_injector_test.go
samuelkarp: That's probably fine; there's a risk that someone who doesn't know that a test requires UID 0 will just miss running them, but since we have this going through CI now I think it's okay.

samuelkarp (approved):

PR #390, opened 2024-07-12
Change for proper behavior of cos gpu driver install

Insufficient-Charge:


Jiaqicao257:
Could you update https://github.com/GoogleCloudPlatform/container-engine-accelerators/blob/master/nvidia-driver-installer/cos/daemonset-preloaded-latest.yaml that installs the latest GPU driver version with the same change as well?

nvidia-driver-installer/cos/daemonset-preloaded-latest.yaml
```diff
@@ -116,8 +116,8 @@ spec:
             echo "GPU driver is already installed, the installed version may or may not be the driver version being tried to install, skipping installation"
             exit 0
           else
-            echo "No GPU driver module detected, installting now"
-            /cos-gpu-installer install --version=latest
+            echo "No GPU driver module detected, installing now"
+            /cos-gpu-installer install --version=latest && exit 0 || exit 1
```
elfinhe, line 120: should we continue running if command succeeds? and better printing an error message as well before exit 1.

            /cos-gpu-installer install --version=latest || exit 1

Jiaqicao257 (approved):

PR #403, opened 2024-08-29
Update restart logic for kubelet restart

grac3gao:
manually tested with kubelet restart


pkg/gpu/nvidia/manager.go
```diff
@@ -489,7 +500,16 @@ func (ngm *nvidiaGPUManager) Serve(pMountPath, kEndpoint, pluginEndpoint string)
 								}
 							}
 						}
-
+					// Restart the device plugin if kubelet socket gets recreated, which indicates a kubelet restart.
```
elfinhe, line 503: How do we test this change?

elfinhe (approved):
LGTM.

pkg/gpu/nvidia/manager.go
grac3gao: Only by manual test. When manually restart the kubelet, the device plugin server will restart as 
```
I0829 21:34:53.142655 1043311 manager.go:506]  /device-plugin/kubelet.sock recreated.
E0829 21:34:53.142767 1043311 manager.go:455] device-plugin server stopped serving: <nil>
```

PR #413, opened 2024-11-04
Add tcpxo daemon anti affinity to avoid duplicate scheduling

thisSIDEofRANDOM:
Adding pod anti affinity rules and a label to examples to reduce the chance of scheduling multiple tcpxo-daemons on a single node.


thisSIDEofRANDOM:
Tested scheduling on two node a3 mega cluster without issue using the affinity rules. Confirmed if the GPU count was removed these rules kept the pods from double booking a single node.

grac3gao (approved):

PR #416, opened 2024-11-18
A3U TAS support

dburl:
- makes A3M GKE scheduler with topology awareness to also work on A3U machines


thisSIDEofRANDOM (dismissed):
Should we split this into its own DIR that is topology scheduling not under any tcpx/tcpxo folder? This solution is meant to be generic so seems like it might be easier.

We could then leave the other files in place

gke-topology-scheduler/schedule-daemon.yaml
```diff
@@ -0,0 +1,39 @@
...
+        effect: "NoSchedule"
+      containers:
+      - name: topology-scheduler-container
+        image: python:3.10
+        command:
+        - /bin/sh
+        - -c
+        - |
+          pip install google-auth cachetools google-api-python-client kubernetes
+          python -u /scripts/schedule-daemon.py --ignored-namespace kube-system gmp-public gmp-system
```
samos123, line 30: as a follow up, could we build and publish an image instead of downloading dependencies at runtime?

samos123:
I would prefer a container image publishes that contains all dependencies and schedule-daemon.py. There will be a time where the dependencies break. 

Should we remove this directory? https://github.com/GoogleCloudPlatform/container-engine-accelerators/tree/master/gpudirect-tcpxo/topology-scheduler

thisSIDEofRANDOM:
> I would prefer a container image publishes that contains all dependencies and schedule-daemon.py. There will be a time where the dependencies break.
> 
> Should we remove this directory? https://github.com/GoogleCloudPlatform/container-engine-accelerators/tree/master/gpudirect-tcpxo/topology-scheduler

I think it would make sense to move the scheduling daemon to its own folder since kueue will be incorporating the same logic. This will leave an example scheduler available while also not implying its limited to one technology. I.e. topology scheduling doesn't really care if you are running tcpxo etc.

gke-topology-scheduler/schedule-daemon.yaml
thisSIDEofRANDOM: I could follow up on building a container shared out. @dburl unless you want to do this as part of the commit.

gke-topology-scheduler/README.md
```diff
@@ -0,0 +1,53 @@
...
+A3M & A3U machines clusters.
+
+The general outline for this to be successful is:
+- We add labels for topology to nodes in the cluster with a daemonset
+- We handle pod scheduling with a scheduling daemon
+- Pods with the added scheduleGate are picked up and scheduled
+
+## Prerequisites
+
+For topology awareness to be enabled, a GKE node pool has to be created with
```
thisSIDEofRANDOM, line 13: I don't believe this is a requirement anymore. DWS doesn't use compact placement but still makes this information available. Do we have a check for this in the code still?

dburl: it might be not a req for GKE >=1.31, let me recheck

dburl: confirmed, that it's not req for A3U

thisSIDEofRANDOM: Thanks, it would be strange for anyone to be installing this on a machine type that doesn't have metadata so hopefully the removal of the check causes no issues

thisSIDEofRANDOM:
Please fix linting and then I think we can approve.

gke-topology-scheduler/schedule-daemon.yaml
dburl: I will add Dockerfile in a separate PR

thisSIDEofRANDOM (approved):
LGTM with the changes. We need to follow up in another PR to remove the old tcpxo specific but lets limit this change for the time being

samos123 (approved):

PR #422, opened 2024-12-11
Add multi-arch support for nvidia-device-plugin

ollyplance:
* update Dockerfile to cross-build on arm
* update Makefile to use buildx with `--platforms` arg

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,10 +12,21 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM golang:1.23-bullseye as builder
+FROM --platform=$BUILDPLATFORM golang:1.23-bullseye AS builder
+
+ARG TARGETOS
+ARG TARGETARCH
+
 WORKDIR /go/src/github.com/GoogleCloudPlatform/container-engine-accelerators
 COPY . .
-RUN go build cmd/nvidia_gpu/nvidia_gpu.go
+RUN if [ "${TARGETARCH}" = "arm64" ] && [ "${BUILDARCH}" != "arm64" ]; then \
+    apt update && \
+    apt install -yq --no-install-recommends \
+        gcc-aarch64-linux-gnu libc6-dev-arm64-cross; \
+        CC=aarch64-linux-gnu-gcc; \
+    fi && \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -12,10 +12,21 @@
...
-FROM golang:1.23-bullseye as builder
+FROM --platform=$BUILDPLATFORM golang:1.23-bullseye AS builder
+
+ARG TARGETOS
+ARG TARGETARCH
+
 WORKDIR /go/src/github.com/GoogleCloudPlatform/container-engine-accelerators
 COPY . .
-RUN go build cmd/nvidia_gpu/nvidia_gpu.go
+RUN if [ "${TARGETARCH}" = "arm64" ] && [ "${BUILDARCH}" != "arm64" ]; then \
```
elfinhe, line 22: can we double check it conditions here? Is it assuming the build machine is on X86?

elfinhe (commented):
LGTM in general.

Dockerfile
```diff
@@ -12,10 +12,21 @@
...
+    apt update && \
+    apt install -yq --no-install-recommends \
+        gcc-aarch64-linux-gnu libc6-dev-arm64-cross; \
+        CC=aarch64-linux-gnu-gcc; \
+    fi && \
```
grac3gao, line 27: This dockerfile is used to internally build image for release, also need to double check if the currently louhi pipeline can work successfully after this update.

ollyplance: Yes, currently the Louhi pipeline runs `make container.` Locally `make container` performs the exact same as before, only using the dockerx wrapper around docker. For actually building the multi-arch build in louhi, I separated out that build into a different make command. I will create a seperate louhi flow to verify that works before proceeding with changing the pipeline.

Dockerfile
ollyplance: Yes the current build machine is x86. These conditions are ensuring that we only install the cross-compiler if the host machine is x86. It may be a good point to ensure that the host machine is nothing other than x86 or arm though, if that is what you were getting at?

grac3gao (approved):

grac3gao:
LGTMed. Please ensure existing path will not be broken.

PR #438, opened 2025-02-11
Fix for duty cycle metrics out of range issue

grac3gao:
We found duty cycle metrics returns out of range [0-100] value, this fix is for resolving this. It also adds more logs and comments for future diagnosis. 

Manually tested.

### DIFF
--- a/pkg/gpu/nvidia/metrics/util.go
+++ b/pkg/gpu/nvidia/metrics/util.go
@@ -34,7 +34,7 @@ package metrics
 // In my experiments, I found that NVML_GPU_UTILIZATION_SAMPLES buffer stores
 // 100 samples that are uniformly spread with ~6 samples per second. So the
 // buffer stores last ~16s of data.
-nvmlReturn_t nvmlDeviceGetAverageUsage(char *uuid, unsigned long long lastSeenTimeStamp, unsigned int* averageUsage) {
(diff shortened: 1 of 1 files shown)


pkg/gpu/nvidia/metrics/util.go
```diff
@@ -83,12 +98,16 @@ func AverageGPUUtilization(uuid string, since time.Duration) (uint, error) {
...
 	C.free(unsafe.Pointer(uuidCStr))
 	if r != C.NVML_SUCCESS {
 		return 0, fmt.Errorf("failed to get GPU utilization for device %s, nvml return code: %v", uuid, r)
-
 	}
-
-	return uint(util), nil
+	averageUtilization := uint(util)
+	if averageUtilization > 100 || averageUtilization < 0 {
+		return 0, fmt.Errorf("failed to get GPU utilization for device %s, out of range [0, 100] utilization: %d, sum: %d, sampleCount: %d", uuid, averageUtilization, uint(sum), uint(sampleCount))
```
elfinhe, line 110: How will the error be handled from the function calller?

grac3gao: It will report empty metric, and log error in the device plugin: https://github.com/GoogleCloudPlatform/container-engine-accelerators/blob/master/pkg/gpu/nvidia/metrics/metrics.go#L196-L199.

PR #450, opened 2025-03-05
Add support for Confidential GPU in the COS GPU Driver Plugin

FigueroaCode:
Add new manifest of COS GPU Driver Plugin with a modified command and node affinity. The installation for confidential GPU has different requirements and thus requires a slightly differently installation procedure. Additionally, this is currently only supported on Confidential TDX Nodes on the a3-highgpu-1g machine types, so a node affinity was added so it is only included on valid nodes.

### DIFF
--- a/gpudirect-rdma/nccl-rdma-installer.yaml
+++ b/gpudirect-rdma/nccl-rdma-installer.yaml
@@ -55,6 +55,22 @@ spec:
           hostPath:
             path: /home/kubernetes/bin/gib
       initContainers:
+        - name: disable-log-martian
+          image: alpine:latest
+          command: ["/bin/sh"]
+          securityContext:
+            privileged: true
+          args:
+          - -c
+          - |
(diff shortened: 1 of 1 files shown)


nvidia-driver-installer/cos/daemonset_confidential.yaml
```diff
@@ -0,0 +1,154 @@
...
+          mountPath: /dev
+        - name: root-mount
+          mountPath: /root
+        - name: cos-tools
+          mountPath: /build/cos-tools
+        command:
+        - bash
+        - -c
+        - |
+          echo "Checking for existing GPU driver modules"
```
grac3gao, line 122: Can we add the node label checking logic in the command too, so that once we need to update the auto driver installation manifest, we can gain more confidence. Not urgent, you can add it in follow up PRs.

FigueroaCode: Yeah no problem. I"ll make a follow up PR with that soon.

PR #470, opened 2025-04-07
Add nvidia persistenced installer binary

FigueroaCode:
For Confidential GPUs persistence mode is required on the GPU to avoid losing the encryption keys. To do this from a container we plan to use a side car container to keep the persistence daemon running on the node.


nvidia-persistenced-installer/nvidia_persistenced_installer.go
```diff
@@ -0,0 +1,167 @@
...
+	}
+	return versionMajor, nil
+}
+
+func checkConfidentialGPUEnablement(ctx context.Context) (bool, error) {
+	file, err := os.ReadFile(*cgpuConfigFile)
+	if err != nil {
+		// Treat non existence of file as disabled.
+		if os.IsNotExist(err) {
+			glog.WarningContextf(ctx, "confidential node type file not found at %v, skipping persistenced installation", *cgpuConfigFile)
```
Jiaqicao257, line 159: why is this warning instead of info? IIUC this log will show up for all non cgpu users, putting it as warning might be confusing.

nvidia-persistenced-installer/nvidia_persistenced_installer.go
```diff
@@ -0,0 +1,167 @@
...
+	} else {
+		glog.InfoContext(ctx, "Confidential GPU is NOT enabled, skipping nvidia persistenced enablement.")
+		// Don't exit as this is intended for a side car which would cause it to restart infinitely.
+	}
+
+	// Need to keep the container running so that the nvidia persistence daemon can keep running.
+	for {
+		glog.InfoContext(ctx, "sleeping");
+		time.Sleep(5 * time.Minute)
+	}
```
Jiaqicao257, line 74: Several questions about this 
1. Can persistend fail after it's started? If so how can it be monitored from this container? 
2. Let's not put a log "sleeping" every 5 mins as it might lead to log spam

nvidia-persistenced-installer/nvidia_persistenced_installer.go
```diff
@@ -0,0 +1,167 @@
...
+	if versionMajor, err := nvidiaVersionMajor(); err != nil {
+		return err
+	} else if versionMajor >= minUVMSupportedVersion {
+		// UVM persistence mode is only available starting at R550.
+		cmdArgs = append(cmdArgs, "--uvm-persistence-mode")
+		glog.InfoContext(ctx, "using --uvm-persistence-mode")
+	}
+	cmdArgs = append(cmdArgs, "--nvidia-cfg-path=" + *containerPathPrefix + "/lib64")
+	persistencedCMD := exec.Command(*containerPathPrefix + "/bin/nvidia-persistenced", cmdArgs...)
+	if err := persistencedCMD.Run(); err != nil {
```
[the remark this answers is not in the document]
FigueroaCode: Yes it only starts the persistence daemon and does not monitor it. So after confirming that it started the command finishes and moves onto the next part.

nvidia-persistenced-installer/nvidia_persistenced_installer.go
FigueroaCode: The persistence daemon after its done just sleeps for the rest of the time. So I think it is unlikely to fail after it has started.

I'm not sure if we have a good way to monitor it, but if it does fail later, from my experience things end up broken even if you try to restart it. It usually just leads to having to reboot the node as you can't get the encryption keys back. So I don't think it's worth it to try to monitor it and restart it.

FigueroaCode: Agreed on the log spam, I'll remove the sleep log.

nvidia-persistenced-installer/nvidia_persistenced_installer.go
```diff
@@ -0,0 +1,167 @@
...
+	}
+	contentStr := string(content)
+	kernelModuleVersionPattern := regexp.MustCompile(`\d+\.\d+\.\d+`)
+	kernelModuleVersion := kernelModuleVersionPattern.FindString(contentStr)
+	glog.Infof("NVIDIA kernel module version: %s", kernelModuleVersion)
+	return kernelModuleVersion
+}
+
+func nvidiaVersionMajor() (int, error) {
+	version := getLoadedNVIDIAKernelModuleVersion("/proc/driver/nvidia/version")
```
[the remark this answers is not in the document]
FigueroaCode: So I tested this to verify my suspicion. Using nvidia-smi at this point breaks things and causes the gpu to get detached, and the only way to recover here is to reboot the node. cos-gpu-installer reads the file directly as well rather than calling nvidia-smi. So I think GPU is not ready for this interaction at this point of the setup so we need to do it this way.

nvidia-persistenced-installer/nvidia_persistenced_installer.go
Jiaqicao257: > this should only be running on confidential GPU nodes

That would be ideal, but I can't think of anyway to realize this for auto installation?

nvidia-persistenced-installer/nvidia_persistenced_installer.go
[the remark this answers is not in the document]
Jiaqicao257: I see, that makes sense, thanks for the verification.

nvidia-persistenced-installer/nvidia_persistenced_installer.go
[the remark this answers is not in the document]
Jiaqicao257: Got it.

nvidia-persistenced-installer/nvidia_persistenced_installer.go
Jiaqicao257: Will the GPU device plugin container fail or report unhealthy GPUs if persistenced fails? I worry it'll just silently fail without the observability to tell what's wrong, then it'll be hard for users and us to debug.

nvidia-persistenced-installer/nvidia_persistenced_installer.go
FigueroaCode: oh thats a good point. Yeah we should keep it as info since it will be used later in the auto installation.

nvidia-persistenced-installer/nvidia_persistenced_installer.go
```diff
@@ -0,0 +1,167 @@
...
+	glog.Infof("Attempting to read version from: %s", versionFilePath)
+	content, err := os.ReadFile(versionFilePath)
+	if err != nil {
+		glog.Infof("Failed to read version file: %v", err)
+		return ""
+	}
+	contentStr := string(content)
+	kernelModuleVersionPattern := regexp.MustCompile(`\d+\.\d+\.\d+`)
+	kernelModuleVersion := kernelModuleVersionPattern.FindString(contentStr)
+	glog.Infof("NVIDIA kernel module version: %s", kernelModuleVersion)
```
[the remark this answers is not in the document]
FigueroaCode: done

nvidia-persistenced-installer/nvidia_persistenced_installer.go
```diff
@@ -0,0 +1,167 @@
...
+	return kernelModuleVersion
+}
+
+func nvidiaVersionMajor() (int, error) {
```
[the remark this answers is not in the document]
FigueroaCode: done

nvidia-persistenced-installer/nvidia_persistenced_installer.go
FigueroaCode: I think it would silently fail. We might be able to monitor it later with the device plugin using the nvidia manager? For example the GPU would get detached and would probably be in an unhealthy state, so that might get picked up by this? https://github.com/GoogleCloudPlatform/container-engine-accelerators/blob/master/cmd/nvidia_gpu/nvidia_gpu.go#L142

nvidia-driver-installer/cos/daemonset-confidential.yaml
```diff
@@ -151,4 +151,4 @@ spec:
           mountPath: /etc/nvidia
       containers:
       - image: "gke.gcr.io/pause:3.8@sha256:880e63f94b145e46f1b1082bb71b85e21f16b99b180b9996407d61240ceb9830"
-        name: pause
+        name: pause
```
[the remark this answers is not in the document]
FigueroaCode: I just changed the file name to be more inline with the other file types. I thought it would be better to be consistent, i.e. use a `-` instead of  `_`. Better to fix this early now before people and tests are using it.

But technically it's not needed

grac3gao (approved):

nvidia-persistenced-installer/nvidia_persistenced_installer.go
Jiaqicao257: Per discussion in chat, I'm fine with merging this as it is now but we probably want to come back later for error detection/handling. Thanks!

Jiaqicao257 (approved):

PR #472, opened 2025-04-16
Add make command to build nvidia persistenced installer image

FigueroaCode:
Add make command for the nvidia persistenced installer image to use it in the automated image pipeline.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,6 +40,7 @@ IMAGE=nvidia-gpu-device-plugin
 PARTITION_GPU_IMAGE=nvidia-partition-gpu
 FASTSOCKET_INSTALLER_IMAGE=fastsocket-installer
 DEVICE_INJECTOR_IMAGE=nri-device-injector
+NVIDIA_PERSISTENCED_IMAGE=nvidia-persistenced-installer
 ALL_ARCHITECTURES=amd64 arm64
 
 build:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -78,6 +79,9 @@ fastsocket_installer:
 nri-device-injector:
 	docker build --pull -t ${REGISTRY}/${DEVICE_INJECTOR_IMAGE}:${TAG} -f nri_device_injector/Dockerfile .
 
+nvidia_persistenced_installer:
```
ollyplance, line 82: nit: can this be changed to `nvidia-persistenced-installer` to keep inline with style of (most) other make targets?

ollyplance (approved):
LGTM

Makefile
ollyplance: Ah no that makes sense! Thank you for the explanation, lets keep it as it is.

ollyplance (approved):

Makefile
```diff
@@ -78,6 +79,9 @@ fastsocket_installer:
...
+	docker buildx build --pull --load -t ${REGISTRY}/${NVIDIA_PERSISTENCED_IMAGE}:${TAG} -f nvidia-persistenced-installer/Dockerfile .
```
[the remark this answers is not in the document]
FigueroaCode: I believe so since we added this in the docker file: https://github.com/GoogleCloudPlatform/container-engine-accelerators/blob/master/nvidia-persistenced-installer/Dockerfile#L22

I wasn't sure if we needed the support for the other architectures in the future but I didn't think it would hurt to have it ready. It's the same setup as the partition_gpu image.

Jiaqicao257 (approved):

PR #474, opened 2025-04-18
Add nvidia-persistenced-installer sidecar to daemonset-confidential.yaml

FigueroaCode:
Adding the nvidia persistenced side car to the gpu driver installer to keep the SPDM session on the GPU. Also removing the validation on the machine type since that is already handled at the patching layer during node pool creation/updates.

### DIFF
--- a/nvidia-driver-installer/cos/daemonset-confidential.yaml
+++ b/nvidia-driver-installer/cos/daemonset-confidential.yaml
@@ -113,8 +113,6 @@ spec:
           mountPath: /build/cos-tools
         - name: nvidia-config
           mountPath: /etc/nvidia
-        - name: nvidia-config
-          mountPath: /etc/nvidia
         command:
         - bash
         - -c
--- a/nvidia-driver-installer/cos/daemonset-confidential.yaml
+++ b/nvidia-driver-installer/cos/daemonset-confidential.yaml
@@ -53,10 +53,6 @@ spec:
                 operator: In
                 values:
                   - TDX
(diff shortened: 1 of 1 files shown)


nvidia-driver-installer/cos/daemonset-confidential.yaml
```diff
@@ -115,24 +111,74 @@ spec:
...
+            value: /home/kubernetes/bin/nvidia/vulkan/icd.d
+          - name: VULKAN_ICD_DIR_CONTAINER
+            value: /etc/vulkan/icd.d
+          - name: ROOT_MOUNT_DIR
+            value: /root
+          - name: COS_TOOLS_DIR_HOST
+            value: /var/lib/cos-tools
+          - name: COS_TOOLS_DIR_CONTAINER
+            value: /build/cos-tools
+        volumeMounts:
```
Jiaqicao257, line 165: Are all these mounts and env vars necessary? shouldn't /usr/local/nvidia be enough?

FigueroaCode: I couldn't completely understand why, but yes they were necessary unfortunately. There would be issues when I took any of them away. I went 1 by 1 to try to minimize the things needed.

Jiaqicao257 (approved):

nvidia-driver-installer/cos/daemonset-confidential.yaml
Jiaqicao257: I see...

PR #487, opened 2025-05-02
Added logic to not destroy and recreate partitions when valid partitions exist. 

rishibathina:
Prior to destroying and recreating partitions, added logic to query current partitions. If the partitions match the desired partitions specified by the partition size and the GPU type, then exit without recreating.

### DIFF
--- a/partition_gpu/partition_gpu.go
+++ b/partition_gpu/partition_gpu.go
@@ -23,6 +23,8 @@ import (
 	"os/exec"
 	"strings"
 	"syscall"
+	"bufio"
+	"regexp"
 
 	"github.com/golang/glog"
 )
@@ -55,6 +57,22 @@ var partitionSizeToProfileID = map[string]string{
 	"3g.71gb":  "9",
 	"4g.71gb":  "5",
 	"7g.141gb": "0",
+
+	//nvidia-b200, nvidia-gb200
+	"1g.23gb": "19",
+	//nvidia-b200
+	"1g.45gb":  "15",
+	"2g.45gb":  "14",
+	"3g.90gb":  "9",
+	"4g.90gb":  "5",
+	"7g.180gb": "0",
+
+	//nvidia-gb200
+	"1g.47gb":  "15",
+	"2g.47gb":  "14",
+	"3g.93gb":  "9",
+	"4g.93gb":  "5",
+	"7g.186gb": "0",
 }
 
 var partitionSizeMaxCount = map[string]int{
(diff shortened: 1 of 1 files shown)


Jiaqicao257:
Please paste the test results here, thanks

rishibathina:
Here is the result of attempting to re-partition:
Screenshot 2025-05-01 at 5 49 37 PM

partition_gpu/partition_gpu.go
```diff
@@ -329,3 +337,102 @@ func buildPartitionStr(partitionSize string) (string, error) {
 
 	return partitionStr, nil
 }
+
+// getCurrentPartitionProfileCounts checks fro active MIG GPU instances,
+// and outputs a map of their profileIDs -> counts
+func checkCurrentPartitionProfileCounts(desiredMaxCount int) bool {
+	args := []string{"mig", "-lgi"}
```
Jiaqicao257, line 344: Parsing the output from a table seems very error prone and tedious, have you tried adding `--format=csv,noheader` to the command to filter out the unnecessary info?

rishibathina: Yes, this command doesn't have csv,no header from my finding.

rishibathina: https://paste.googleplex.com/5352794967769088

testing on a3

cmd/nvidia_gpu/nvidia_gpu
Jiaqicao257, line 1: Please remove this.

partition_gpu/partition_gpu.go
Jiaqicao257, line 1: Could you add some unit tests for the methods you added?

Jiaqicao257 (approved):

grac3gao:
error 

```
Run make presubmit
>> vetting code
>> checking go formatting
The following files are not properly formatted:
partition_gpu/partition_gpu_test.go
make: *** [Makefile:33: presubmit] Error 1
```
please fix

partition_gpu/partition_gpu.go
```diff
@@ -329,3 +335,100 @@ func buildPartitionStr(partitionSize string) (string, error) {
...
+	if !uniform {
+		return uniform
+	}
+	desiredPartition := checkDesired(partitions, desiredMaxCount)
+
+	return desiredPartition
+}
+
+// figure out return type
+func parseLGIOutput(lgiOutput string) (map[string][]string, bool) {
```
grac3gao: In this function, when `bool` is false, it has a couple of possible reasons, can we directly return `err` with some specific error message, and replace line351 ~ 352 as:
```
if err != nil {
glog.Errorf("failed to parseLGIOutput: %v", err)
return false
}
```
so that when parseLGIOutput return false, we know which specific part goes wrong.

partition_gpu/partition_gpu.go
```diff
@@ -329,3 +335,103 @@ func buildPartitionStr(partitionSize string) (string, error) {
...
+		return false
+	}
+	glog.Infof("Output:\n %s", string(out))
+	outputText := string(out)
+	partitions, uniform, lgiError := parseLGIOutput(outputText)
+	if lgiError != nil {
+		glog.Errorf("failed to parse 'nvidia-smi mig -lgi' output: %v", lgiError)
+		return false
+	}
+	if !uniform {
```
grac3gao: add a log when it return false with no err. Or non-uniform case can also be considered as a error case?

grac3gao:
`Merging is blocked
Commits must have verified signatures.`

Please fix it.

grac3gao (dismissed):

grac3gao (approved):

PR #497, opened 2025-06-23
Add XidCriticalError in condition

sgysherry:
1. Extend the XID range(31, 45), only for test purpose
2. Patch the gpu-device-plugin
3. Trigger XID, observe the condition get applied
```
k describe node gke-weitong-cluster-p100-22d24db3-kd6j | grep Xid
  XidCriticalError     True    Mon, 23 Jun 2025 18:45:43 +0000   Mon, 23 Jun 2025 18:44:57 +0000   {"31":true}                  6e43a052-e544-467b-a972-63d7cd2ce80a
```
5. Delete the XID workflow (simulate the case that hardware XID is only thrown out once), the condition persists, and the heartbeat time get updated
```
  - lastHeartbeatTime: "2025-06-23T18:58:44Z"
    lastTransitionTime: "2025-06-23T18:44:57Z"
    message: 6e43a052-e544-467b-a972-63d7cd2ce80a
    reason: '{"31":true}'
    status: "True"
    type: XidCriticalError
```
7. Restart the gpu-device-plugin, the condition persists
8. Do VM Restart on the node, the condition get deleted

Safe error failback:  
Any of the failure in the new added feature will not block the previous workflow. 

Another PR will be generated for metrics(k8s node event)


demo/gpu-error/illegal-memory-access/vectorAdd.cu
```diff
@@ -21,7 +21,7 @@ limitations under the License.
 
  #include <stdio.h>
  #include <cuda_runtime.h>
- #include <helper_cuda.h>
+//  #include <helper_cuda.h>
```
SidneyShen: Remove?

pkg/gpu/nvidia/health_check/health_checker.go
```diff
@@ -55,13 +74,66 @@ func NewGPUHealthChecker(devices map[string]pluginapi.Device, health chan plugin
 		glog.Infof("reading code %v", c)
 		hc.healthCriticalXid[uint64(c)] = true
 	}
+
+	monitorCriticalXid := []int{48, 63, 64, 79, 119, 120, 123, 140}
```
grac3gao, line 78: Will there be new Xid added in the future or customer want to monitor a specific Xid which doesn't include here? Any way we can do to avoid updating codes every time when new XID need to be added? e.g. through configmap like healthCriticalXid does?

pkg/gpu/nvidia/health_check/health_checker.go
```diff
@@ -55,13 +74,66 @@ func NewGPUHealthChecker(devices map[string]pluginapi.Device, health chan plugin
...
+		}
+	} else {
+		glog.Infof("XIDCriticalError condition doesn't exist for node %s.", hc.nodeName)
+	}
+	return nil
+}
+
 // Start registers NVML events and starts listening to them
 func (hc *GPUHealthChecker) Start() error {
+	nodeName, err := metadata.InstanceNameWithContext(context.Background())
```
grac3gao, line 126: any permission needed to call this metadata API?

rbac.yaml
```diff
@@ -0,0 +1,53 @@
+# Copyright 2018 Google Inc. All Rights Reserved.
```
grac3gao: consdier move this file under /cmd/nvidi_gpu

pkg/gpu/nvidia/health_check/health_checker.go
```diff
@@ -55,13 +74,66 @@ func NewGPUHealthChecker(devices map[string]pluginapi.Device, health chan plugin
...
+func (hc *GPUHealthChecker) resetXIDCondition() error {
+	node, err := hc.kubeClient.CoreV1().Nodes().Get(context.Background(), hc.nodeName, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+	bootId := node.Status.NodeInfo.BootID
+	lastBootId := ""
+	newConditions := []v1.NodeCondition{}
+	for _, condition := range node.Status.Conditions {
+		if condition.Type == XIDConditionType && condition.Status == "True" {
```
grac3gao, line 100: Is there any case that the node has XIDcondition while status is false?

pkg/gpu/nvidia/health_check/health_checker.go
```diff
@@ -176,12 +248,107 @@ func (gd *GPUDevice) parseMigDeviceUUID(UUID string) (string, uint, uint, error)
...
+
+func (hc *GPUHealthChecker) setXIDheartbeat() {
+	for {
+		select {
+		case <-hc.stop:
+			return
+		default:
+			hc.updateLastHeartbeatTime()
+			time.Sleep(1 * time.Minute)
+		}
```
grac3gao, line 319: Does this mean once we have XID, no matter it disappeared or not, the heartbeat will always be updated, until the node got repaired?

sgysherry: Yes. Condition normally has a LastHeartbeatTime to represent its latest status(best practice)), but the critical XID is non-recoverable. It will update the heartbeat all the time until some actions like reboot(what we are doing now), or some other GKE customized recovery behavior(not implemented) get performed.

pkg/gpu/nvidia/health_check/health_checker.go
sgysherry: Not any I can see now. It will either be true or non-exist.

pkg/gpu/nvidia/health_check/health_checker.go
sgysherry: Internally it is calling the metadata service with `getETag` at

`"http://" + host + "/computeMetadata/v1/" + suffix`

I don't think any special permission is needed

pkg/gpu/nvidia/health_check/health_checker.go
```diff
@@ -55,13 +74,66 @@ func NewGPUHealthChecker(devices map[string]pluginapi.Device, health chan plugin
...
+	} else {
+		glog.Infof("XIDCriticalError condition doesn't exist for node %s.", hc.nodeName)
+	}
+	return nil
+}
+
 // Start registers NVML events and starts listening to them
 func (hc *GPUHealthChecker) Start() error {
+	nodeName, err := metadata.InstanceNameWithContext(context.Background())
+	if err != nil {
```
SidneyShen, line 127: Will this fail the device plugin main process?

pkg/gpu/nvidia/health_check/health_checker.go
sgysherry: Accidentally deleted xinnings' comment:

These xid error codes are selected from internal doc based on repairbility of xid errors. In theory each xid errors need to be used together with other conditions to determine criticalness., like companion with other xids, or certain GPU types. We are not expecting to update this list quite often. Current solution is just surfacing the critical xid codes that GKE determined (similar to xid 48) through node condition, and customer could use that to trigger repair through GPU reboot/reset API. If customer wants to instruct GKE to repair certain xids, I think current cnofigMap interface should be sufficient for that purpose.

pkg/gpu/nvidia/health_check/health_checker.go
sgysherry: Yes, updated it to be error logging only

SidneyShen (approved):

grac3gao (approved):

PR #499, opened 2025-06-27
Add XID event

sgysherry:
```
I0628 00:10:03.285992 1201124 event.go:298] Event(v1.ObjectReference{Kind:"Node", Namespace:"", Name:"gke-weitong-cluster-p100-22d24db3-kd6j", UID:"3bc1f4fe-3e27-445d-a3f7-9ef97d2c0acd", APIVersion:"v1", ResourceVersion:"1751069402089711001", FieldPath:""}): type: 'Warning' reason: 'XIDError' Catch XID error, XID=31
```
image

Also tested with same flow like https://github.com/GoogleCloudPlatform/container-engine-accelerators/pull/497


pkg/gpu/nvidia/health_check/health_checker.go
```diff
@@ -340,13 +351,31 @@ func (hc *GPUHealthChecker) updateLastHeartbeatTime() {
 	}
 }
 
+func (hc *GPUHealthChecker) recordXIDEvent(e nvml.Event) error {
+	event := &v1.Event{
+		Type:    v1.EventTypeWarning,
+		Reason:  "XIDError",
+		Message: "Catch XID error, XID=" + strconv.FormatUint(e.Edata, 10),
+		Source: v1.EventSource{
+			Component: "gpu-device-plugin:health_checker",
```
linxiulei: is gpu-device-plugin sufficient here?

pkg/gpu/nvidia/health_check/health_checker.go
```diff
@@ -340,13 +351,31 @@ func (hc *GPUHealthChecker) updateLastHeartbeatTime() {
 	}
 }
 
+func (hc *GPUHealthChecker) recordXIDEvent(e nvml.Event) error {
+	event := &v1.Event{
```
linxiulei: it doesn't need to create event object here but uses recorder.Eventf() in the below

pkg/gpu/nvidia/health_check/health_checker.go
```diff
@@ -340,13 +351,31 @@ func (hc *GPUHealthChecker) updateLastHeartbeatTime() {
...
+		Type:    v1.EventTypeWarning,
+		Reason:  "XIDError",
+		Message: "Catch XID error, XID=" + strconv.FormatUint(e.Edata, 10),
```
linxiulei: nit: past tense?

pkg/gpu/nvidia/health_check/health_checker.go
```diff
@@ -340,13 +351,31 @@ func (hc *GPUHealthChecker) updateLastHeartbeatTime() {
...
+		Source: v1.EventSource{
+			Component: "gpu-device-plugin:health_checker",
+		},
+	}
+	node, err := hc.kubeClient.CoreV1().Nodes().Get(context.Background(), hc.nodeName, metav1.GetOptions{})
```
linxiulei: we don't need latest node object for event emitting purpose and getting node from etcd would result in visible overhead to kubernetes apiserver/etcd if volume is high.

Can we get the node from informer's lister or some local cache?

pkg/gpu/nvidia/health_check/health_checker.go
sgysherry: Will fix in next commit

pkg/gpu/nvidia/health_check/health_checker.go
```diff
@@ -128,6 +144,18 @@ func (hc *GPUHealthChecker) Start() error {
...
+	informerFactory := informers.NewSharedInformerFactoryWithOptions(
+		hc.kubeClient,
+		0,
+		informers.WithTweakListOptions(func(options *metav1.ListOptions) {
+			options.FieldSelector = fields.OneTermEqualSelector("metadata.name", hc.nodeName).String()
+		}),
+	)
+	ctx := context.Background()
+	hc.nodeLister = informerFactory.Core().V1().Nodes().Lister()
+	informerFactory.Start(ctx.Done())
```
linxiulei: Do we need to shutdown this in Stop()

sgysherry: Wanna make sure I understand correctly: informerFactory is based on context channel now. Are we suggesting that since informerFactory is part of healthChecker, this should be bounded with the hc.stop instead of context?

SidneyShen: I'm a bit surprised that the [main()](https://github.com/GoogleCloudPlatform/container-engine-accelerators/blob/8e3b5b6df9e77182e9bb367a73fa688eb8366e24/cmd/nvidia_gpu/nvidia_gpu.go#L74) does not create the root context, and you create two separate contexts in this function. The health_checker embeds a stop channel, which I think is unnecessary if we have the context passed from main().

grac3gao: Also for the eventBroadcaster, do we need to do things like 
```
watches := []watch.Interface{
	eventBroadcaster.StartLogging(glog.Infof)
	eventBroadcaster.StartRecordingToSink(&clientv1.EventSinkImpl{Interface: hc.kubeClient.CoreV1().Events("")})
}

go func() {
	<-ctx.Done()
	for _, w := range watches {
		w.Stop()
	}
}()
```
which stop the event watchers when the context is cancelled.

linxiulei: > Wanna make sure I understand correctly: informerFactory is based on context channel now. Are we suggesting that since informerFactory is part of healthChecker, this should be bounded with the hc.stop instead of context?

I believe you can either cancel the context or shutdown the factory explicitly to release resources.

sgysherry: Two problems described in this comment:
1. Explicitly close the broadcaster and informerFactory (Added in the stop function)
2. Context best practice (leave for further discussion, if needed possibly another pr)

pkg/gpu/nvidia/health_check/health_checker.go
```diff
@@ -27,15 +27,23 @@ import (
...
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	client "k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/record"
 
 	pluginapi "k8s.io/kubelet/pkg/apis/deviceplugin/v1beta1"
 )
 
 const (
 	XIDConditionType = "XidCriticalError"
+	eventSource      = "gpu-device-plugin"
```
Jiaqicao257: nvidia-gpu-device-plugin to stay aligned with actual component/DS name

pkg/gpu/nvidia/health_check/health_checker.go
```diff
@@ -340,13 +370,23 @@ func (hc *GPUHealthChecker) updateLastHeartbeatTime() {
 	}
 }
 
+func (hc *GPUHealthChecker) recordXIDEvent(e nvml.Event) error {
+	node, err := hc.nodeLister.Get(hc.nodeName)
+	if err != nil {
+		return err
```
Jiaqicao257, line 376: Could you log the error instead if it's no-op? The call in catchError doesn't really do anything to this err

PR #503, opened 2025-07-02
Add driver installation support and update persistenced installer for cGPU

grac3gao:
add driver installation manifest
update persistenced installer

manually tested with the new image


nvidia-driver-installer/cos/daemonset-confidential-latest.yaml
```diff
@@ -0,0 +1,183 @@
...
+        - name: nvidia-install-dir-host
+          mountPath: /usr/local/nvidia
+        - name: root-mount
+          mountPath: /root
+        - name: nvidia-config
+          mountPath: /etc/nvidia
+        - name: dev
+          mountPath: /dev
+      - image: "gcr.io/gke-release/nvidia-partition-gpu@sha256:116be6b7335c1d34366223b9a3780fe80d862fcf06cd2c580426fdc1697af693"
+        name: partition-gpus
```
Jiaqicao257: Is partition GPU supported for cgpu?

nvidia-persistenced-installer/nvidia_persistenced_installer.go
```diff
@@ -64,8 +65,18 @@ func main() {
 		// Add small delay before setting the ready state for consistency.
 		// If the workload starts too close to when the persistence daemon has started sometimes there can be errors.
 		time.Sleep(time.Duration(*readyDelay) * time.Millisecond)
-		if err := setGPUReadyState(ctx); err != nil {
-			glog.ExitContextf(ctx, "failed to set gpu to ready state: %v", err)
+
+		// Mark GPU with ready state. If encounter no devices were found issue, reboot the node.
+		if output, err := setGPUReadyState(ctx); err != nil {
```
Jiaqicao257, line 70: Can we recommend to disable KCP auto upgrade in user guide till this is fixed? An unexpected node reboot is pretty bad. Alternatively, is it possible to check in this container if persistenced is already running (in case of KCP upgrade) and skip this container if so?

grac3gao: > Can we recommend to disable KCP auto upgrade in user guide till this is fixed? 

Yes, we need clear documentation for this limitation, for KCP upgrade will make node reboot, and pod termination. 

> Alternatively, is it possible to check in this container if persistenced is already running (in case of KCP upgrade) and skip this container if so?

This is not working. persistenced is only running when this container is running. For a KCP upgrade, this container will be terminated, and persistenced also stoped, once persistenced stopped, the problem happened. In cGPU VM, skip this container will also cause problem (say just install the driver then initialize device plugin), device plugin will stuck, no GPU pods can be scheduled. Rebooting the node when we see `no device found `error  is the only workable way for now to let the cGPU + device plugin work properly.

nvidia-driver-installer/cos/daemonset-confidential-latest.yaml
grac3gao: No, removed this.

nvidia-persistenced-installer/nvidia_persistenced_installer.go
Jiaqicao257: I see, thanks for the clarification

Jiaqicao257 (approved):

PR #504, opened 2025-07-08
[A4X] Update NCCL installer and test image version

chiayi:
Update NCCL installer and test from 1.0.4 to 1.0.6.


gpudirect-rdma/nccl-rdma-installer-a4x.yaml
Jiaqicao257, line 1: Also update https://github.com/GoogleCloudPlatform/container-engine-accelerators/blob/master/gpudirect-rdma/nccl-test-a4x.yaml?

Jiaqicao257: Actually let's remove https://github.com/GoogleCloudPlatform/container-engine-accelerators/blob/master/gpudirect-rdma/nccl-test-a4x.yaml since it's only for preview.

chiayi: Done!

PR #514, opened 2025-08-19
Using hybrid method to combine allocatable wait and flock wait

zxqlxy:
Add the local lock to avoid deadlock when we enable the maxSurge rollingUpdate strategy as there will be two gpu-device-plugin running at the same time during the rollingUpdate. Use remote node status when there is no lock there to prevent deadlock though a little bit slower.

### DIFF
--- a/cmd/nvidia_gpu/nvidia_gpu.go
+++ b/cmd/nvidia_gpu/nvidia_gpu.go
@@ -15,10 +15,12 @@
 package main
 
 import (
+	"context"
 	"encoding/json"
 	"flag"
 	"fmt"
 	"io/ioutil"
+	"os"
 	"time"
 
 	gpumanager "github.com/GoogleCloudPlatform/container-engine-accelerators/pkg/gpu/nvidia"
@@ -27,6 +29,8 @@ import (
 	util "github.com/GoogleCloudPlatform/container-engine-accelerators/pkg/gpu/nvidia/util"
 	"github.com/NVIDIA/go-nvml/pkg/nvml"
 	"github.com/golang/glog"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/watch"
 	pluginapi "k8s.io/kubelet/pkg/apis/deviceplugin/v1beta1"
 )
 
@@ -35,6 +39,9 @@ const (
 	kubeletEndpoint      = "kubelet.sock"
 	pluginEndpointPrefix = "nvidiaGPU"
 	devDirectory         = "/dev"
+	nodeNameEnv          = "NODE_NAME"
+	lockFilePath         = "/device-plugin/tpu-device-plugin.lock"
+
 	// Proc directory is used to lookup the access files for each GPU partition.
 	procDirectory = "/proc"
 )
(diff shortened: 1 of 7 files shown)


pkg/gpu/nvidia/util/util.go
```diff
@@ -60,11 +70,105 @@ func BuildKubeClient() (client.Interface, error) {
...
+	kubeClient, err := client.NewForConfig(config)
 	if err != nil {
 		glog.Errorf("failed to get kube client. Error: %v", err)
 		return nil, err
 	}
 
 	return kubeClient, nil
 }
+
+func GetEnvName(envName string) (string, error) {
```
linxiulei: nit: GetEnv?

pkg/gpu/nvidia/util/util.go
```diff
@@ -60,11 +70,105 @@ func BuildKubeClient() (client.Interface, error) {
 	}
 	config.ContentType = runtime.ContentTypeProtobuf
 
-	kubeClient, err := kubernetes.NewForConfig(config)
```
linxiulei, line 63: Seems unnecessary change

cmd/nvidia_gpu/nvidia_gpu.go
```diff
@@ -151,6 +161,29 @@ func main() {
...
+			glog.Infof("Failed to build kube client: %v", err)
+			return
+		}
+		nodeName, err := util.GetEnvName(nodeNameEnv)
+		if err != nil {
+			glog.Infof("Failed to get node name from environment variable %s: %v", nodeNameEnv, err)
+			return
+		}
+		watchfunc := func(options metav1.ListOptions) (watch.Interface, error) {
+			return kubeClient.CoreV1().Nodes().Watch(context.TODO(), metav1.ListOptions{
```
linxiulei: should we reuse `ctx` instead of creating a new context?

pkg/gpu/nvidia/util/util.go
zxqlxy: This is because L24-25 are duplicates.

cmd/nvidia_gpu/nvidia_gpu.go
zxqlxy: fixed in the newest commit

pkg/gpu/nvidia/util/util.go
zxqlxy: Use the same name as in tpu-device-plugin

pkg/gpu/nvidia/util/util.go
linxiulei: you don't need to rename it in import, which was what I meant

pkg/gpu/nvidia/util/util.go
linxiulei: Let's do better this time

cmd/nvidia_gpu/nvidia_gpu.go
```diff
@@ -151,6 +161,29 @@ func main() {
...
+
+	}
+
+	if *enableFlockWait {
+		kubeClient, err := util.BuildKubeClient()
+		if err != nil {
+			glog.Infof("Failed to build kube client: %v", err)
+			return
+		}
+		nodeName, err := util.GetEnv(nodeNameEnv)
```
linxiulei, line 173: This would require a change in DaemonSet yaml, right? should we getHostname if failed to get this env var?

PR #517, opened 2025-08-21
Release 0821

kugurst:


gpudirect-tcpxo/README.md
```diff
@@ -51,6 +52,90 @@ For best practices, refer to [Best practice to run workload with GPUDirect-TCPX(
...
+"NCCL_NET_GDR_LEVEL=PIX",
+"NCCL_FASTRAK_ENABLE_HOTPATH_LOGGING=0",
+"NCCL_FASTRAK_USE_LLCM=1",
+"NCCL_DEBUG=WARN",
+```
+#### What is new with in release:
+* This release coincides with the open-source release of GPUDirect-TCPXO: [https://github.com/google/nccl-plugin-gpudirect-tcpxo](https://github.com/google/nccl-plugin-gpudirect-tcpxo)  
+  * The docker images provided are built from the open-source versions of RxDM and the NCCL network plugin.
+  * These open-source based image are **drop-in** **replacements** for the previous closed-source releases.
+* As of this release, TCPX (for A3 High machines) components are no longer included in the provided Docker images. Please reference the corresponding TCPX documentation for deploying and using those components:
```
Jiaqicao257: we don't really use the same images for TCPX and TCPXO for GKE, instead we use `us-docker.pkg.dev/gce-ai-infra/gpudirect-tcpx/nccl-plugin-gpudirecttcpx-dev:v3.1.9` and `us-docker.pkg.dev/gce-ai-infra/gpudirect-tcpx/tcpgpudmarxd-dev:v2.0.12`. Can we remove this entry to avoid confusion?

kugurst: Done

gpudirect-tcpxo/README.md
```diff
@@ -51,6 +52,88 @@ For best practices, refer to [Best practice to run workload with GPUDirect-TCPX(
...
+    NCCL_NVLSTREE_MAX_CHUNKSIZE=131072
+    # The following have been removed:
+    # NCCL_NVLS_ENABLE=0
+    # NCCL_ALGO=Ring,Tree
+    ```
+
+  * The guest config checker **will enforce** that `NCCL_NVLSTREE_MAX_CHUNKSIZE` is set to this value and will abort any workload where it is not set.  
+  * The presence of `NCCL_NVLS_ENABLE` or `NCCL_ALGO` will only be flagged as unexpected environment variables.  
+    * Moving forward, we recommend that `NCCL_ALGO` remains unset to allow NCCL core and the tuner plugin to select the most appropriate algorithm.  
+* As with previous releases, customers may still use the corresponding `simple` profile (`nccl-env-profile-simple.sh`) and configuration files which do not leverage LL128/NVLSTree (`a3plus_tuner_config_simple.textproto` and `a3plus_guest_config_simple.textproto`)
```
Jiaqicao257: Does this need to be visible to users? what's the recommendation to avoid this?

kugurst: Yea, the environment variables are part of user's deployments, so they need to be aware of the fact that the new default changes what is the valid configuration.

If they don't want to use NVLSTree (we recommend they do), then the recommendation is to use the `-simple.sh` profile script instead of the normal one.

Actually, I should probably update the guidance in [best-practice.md](https://github.com/GoogleCloudPlatform/container-engine-accelerators/blob/master/gpudirect-tcpxo/best-practice.md)

Jiaqicao257: Could you update the wording to be something like " if you want to leverage LL128/NVLSTree (`a3plus_tuner_config_simple.textproto` and `a3plus_guest_config_simple.textproto`), it's recommended to use nccl-env-profile.sh instead of nccl-env-profile-simple.sh"? The current wording is a bit confusing since it's on its own bullet point, or did you mean to add it as a sub-point for "Following the v1.0.11 release which enabled NVLSTree support, this release now incorporates `NVLSTree` into the default configuration for the environment setup."?

kugurst: Yea, the latter is the interpretation I was going for. For full context, the current state of the profile scripts:

- `nccl-env-profile.sh` enables NVLSTree and LL128.
  - In the previous release, it just enabled LL128.
- `nccl-env-profile-simple.sh` does not enable NVLSTree nor LL128.
- `nccl-env-profile-ll128.sh` and `nccl-env-profile-nvlstree.sh` are now just symlinks to `nccl-env-profile.sh`.

PR #542, opened 2025-11-18
Add multi-arch support for nri-device-injector

pwschuurman:
## What does this PR do?

Adds a `nri-device-injector-multi-arch` target for building amd64 & arm64 image targets for the `nri_device_injector`

## Tests

Validated build & push with:

```
_ADDONS_CONTAINER_REGISTRY=gcr.io/psch-gke-dev
_REPO_NAME=nri-device-injector
_TAG=v2

make nri-device-injector-multi-arch REGISTRY=${_ADDONS_CONTAINER_REGISTRY} DEVICE_INJECTOR_IMAGE=${_REPO_NAME} TAG=${_TAG}
make push-all REGISTRY=${_ADDONS_CONTAINER_REGISTRY} IMAGE=${_REPO_NAME} TAG=${_TAG}
make push-multi-arch REGISTRY=${_ADDONS_CONTAINER_REGISTRY} IMAGE=${_REPO_NAME} TAG=${_TAG}
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -79,6 +79,11 @@ fastsocket_installer:
 nri-device-injector:
 	docker build --pull -t ${REGISTRY}/${DEVICE_INJECTOR_IMAGE}:${TAG} -f nri_device_injector/Dockerfile .
 
+nri-device-injector-multi-arch:
+	@for arch in $(ALL_ARCHITECTURES); do \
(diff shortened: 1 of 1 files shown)


Makefile
Jiaqicao257, line 1: Does the dockerfile need to be updated similar to https://github.com/GoogleCloudPlatform/container-engine-accelerators/commit/9837c271e118335d3d572553e4182f939ea82bba?

pwschuurman: I don't think so, this compiled fine for both architectures.

The nri-device-injector does not import any cgo code, so I think the standard `buildx build` should work.

PR #543, opened 2025-11-19
fix(asapd): Restart asapd-lite on dadfailed and retry

gauravkghildiyal:
CC: @Jiaqicao257 
CC: @Weil0ng

### DIFF
--- a/asapd-lite-installer/asapd-lite-installer-a4x-max-bm-cos.yaml
+++ b/asapd-lite-installer/asapd-lite-installer-a4x-max-bm-cos.yaml
@@ -43,6 +43,7 @@ spec:
         command: ["/bin/bash", "-c"]
         args:
           - |
+            set -x
             echo "Copying ipam binary to host persistent storage..."
             cp -f /usr/local/sbin/ipam /var/lib/google/asapd-lite/ipam
             if [ $? -ne 0 ]; then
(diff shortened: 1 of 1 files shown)


asapd-lite-installer/asapd-lite-installer-a4x-max-bm-cos.yaml
```diff
@@ -43,14 +43,39 @@ spec:
...
-            exec /asapd-lite/controller/asapd-lite-controller
+            /asapd-lite/controller/asapd-lite-controller &
+
+            # Wait for a bit to let interfaces come up.
+            sleep 15
+
+            DADFAILED_FOUND=0
+            INTERFACES=$(ip -o link show | awk -F': ' '{print $2}' | grep 'gpu')
+
+            for iface in $INTERFACES; do
```
Jiaqicao257, line 62: Should there be any timeout logic here to prevent it from a infinite loop? I know during testing the dadfailed issue was fixed on one trial normally, but is there any possibility of stuck in the loop?

gauravkghildiyal: So if it's stuck in a loop, then it's anyways not going to work right. I imagine that the container crashlooping atleast makes the failure more apparent

Jiaqicao257: the container will crashlooping because `/asapd-lite/controller/asapd-lite-controller` can't run?

Jiaqicao257: Synced offline, this LGTM

Jiaqicao257 (approved):

PR #547, opened 2025-11-26
Update nri-device-injector image version

pwschuurman:
Update nri-device-injector image has. This is a newer multiarch image that is built from https://github.com/GoogleCloudPlatform/container-engine-accelerators/pull/546

### DIFF
--- a/nri_device_injector/nri-device-injector-autopilot.yaml
+++ b/nri_device_injector/nri-device-injector-autopilot.yaml
@@ -12,9 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# This daemonset installs nvidia driver 450.80.02 and invokes the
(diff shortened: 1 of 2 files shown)


nri_device_injector/nri-device-injector-autopilot.yaml
```diff
@@ -78,7 +78,7 @@ spec:
               systemctl restart containerd.service
             fi
       containers:
```
Jiaqicao257, line 48: Also update the node affinity in this file to land on all accelerator nodes?

Jiaqicao257 (approved):

nri_device_injector/nri-device-injector-autopilot.yaml
pwschuurman: Updated to just GPU nodes (and updated description to reflect the current behavior).

PR #550, opened 2025-12-12
Change A4X-Max NCCL test to use DRANet

ankitjoju-google:


gpudirect-rdma/nccl-test-a4x-max.yaml
```diff
@@ -26,134 +8,110 @@ spec:
     resourceClaimTemplate:
       name: nccl-test-compute-domain-channel
 ---
-apiVersion: v1
-kind: Pod
+apiVersion: resource.k8s.io/v1
```
gauravkghildiyal, line 29: While the new YAML involving stateful set should also work, I think for now we'd prefer to minimize changes and focus solely on:

* Add a ResourceClaimTemplate for mrdma devices (renamed from all-mrdma to `nccl-test-all-mrdma`
* For both pods, replace `privileged: true` with `securityContext: {capabilities: {add: ["IPC_LOCK"]}}` 
* Add a resource claim for rdma to both pods using the nccl-test-all-mrdma template

gauravkghildiyal:
Thank you! 
LGTM

Jiaqicao257 (approved):

PR #551, opened 2025-12-13
fix(asapd): Make asapd-lite daemonset more robust

srajashekara:
- Use a wrapper script to run asapd-lite to handle restarts and failures gracefully.
- Add hostPID to allow debugging and interaction with host processes.
- Add mounts for systemd and fix udev rules path to allow better integration with the host.

### DIFF
--- a/asapd-lite-installer/asapd-lite-installer-a4x-max-bm-cos.yaml
+++ b/asapd-lite-installer/asapd-lite-installer-a4x-max-bm-cos.yaml
@@ -41,7 +41,7 @@ spec:
       hostPID: true
       containers:
       - name: asapd-lite
-        image: us-docker.pkg.dev/gce-ai-infra/asapd-lite/asapd-lite:v0.0.1
+        image: us-docker.pkg.dev/gce-ai-infra/asapd-lite/asapd-lite:v0.0.2
 
         command: ["/bin/bash", "-c"]
         args:


asapd-lite-installer/asapd-lite-installer-a4x-max-bm-cos.yaml
```diff
@@ -36,6 +36,9 @@ spec:
       # host's CX PFs.
       hostNetwork: true
 
+      # Use the host's PID namespace. This can be useful for
+      # debugging and interacting with host processes.
+      hostPID: true
       containers:
       - name: asapd-lite
         image: us-docker.pkg.dev/gce-ai-infra/asapd-lite/asapd-lite:v0.0.1
```
gauravkghildiyal: Don't we need to change the image?

srajashekara: I updated it to 0.0.2.

Jiaqicao257 (approved):

asapd-lite-installer/asapd-lite-installer-a4x-max-bm-cos.yaml
```diff
@@ -36,22 +36,18 @@ spec:
       # host's CX PFs.
       hostNetwork: true
 
+      # Use the host's PID namespace. This can be useful for
+      # debugging and interacting with host processes.
+      hostPID: true
```
pwschuurman, line 41: Do we need hostPID?

srajashekara: In the IPvlan mode we run a ipam as a systemd task on the host and this systemd task is controlled (installed and started) from within the asapd-lite pod. For that we need hostPID

PR #579, opened 2026-03-11
emit nvidia driver version to node annotations

leiyiz:
This change publishes nvidia gpu driver version as node annotation
```
  ...
Annotations:        cloud.google.com/cuda.driver-version.full: 580.105.08
                    cloud.google.com/cuda.driver-version.major: 580
                    cloud.google.com/cuda.driver-version.minor: 105
                    cloud.google.com/cuda.driver-version.revision: 08
                    container.googleapis.com/instance_id: 1343450457406822318
  ...
```

Requires `NODE_NAME` env var set appropriately and `patch` added to rbac rules:
```yml
- name: NODE_NAME
  valueFrom:
    fieldRef:
      fieldPath: spec.nodeName
```
```yml
rules:
- apiGroups: [""]
  resources: ["nodes"]
  verbs: ["get", "list", "watch", "patch"]
```
The patching logic is implemented in a non-blocking manner; it runs in a separate goroutine, ensuring that even if the kube client fails to initialize, the node name is missing, or the annotation patching process encounters an error, the main GPU device plugin functionality remains unaffected and continues to serve normally.

## Tested
Nodepool upgrade with auto install:  
new nodes are created, behaves normally as fresh start.

Manual installation - default -> manual installation - latest -> `sudo reboot`:  
The default versions are correctly emitted. After the reboot, the new latest versions are correctly emitted.

### DIFF
--- a/cmd/nvidia_gpu/nvidia_gpu.go
+++ b/cmd/nvidia_gpu/nvidia_gpu.go
@@ -15,16 +15,19 @@
 package main
 
 import (
+	"context"
 	"encoding/json"
 	"flag"
 	"fmt"
 	"io/ioutil"
+	"os"
 	"time"
 
 	gpumanager "github.com/GoogleCloudPlatform/container-engine-accelerators/pkg/gpu/nvidia"
 	healthcheck "github.com/GoogleCloudPlatform/container-engine-accelerators/pkg/gpu/nvidia/health_check"
 	"github.com/GoogleCloudPlatform/container-engine-accelerators/pkg/gpu/nvidia/metrics"
 	util "github.com/GoogleCloudPlatform/container-engine-accelerators/pkg/gpu/nvidia/util"
+	versionvisibility "github.com/GoogleCloudPlatform/container-engine-accelerators/pkg/gpu/nvidia/version_visibility"
 	"github.com/NVIDIA/go-nvml/pkg/nvml"
 	"github.com/golang/glog"
 	pluginapi "k8s.io/kubelet/pkg/apis/deviceplugin/v1beta1"
@@ -50,6 +53,7 @@ var (
 	gpuMetricsPort                 = flag.Int("gpu-metrics-port", 2112, "Port on which GPU metrics for containers are exposed")
 	gpuMetricsCollectionIntervalMs = flag.Int("gpu-metrics-collection-interval", 30000, "Collection interval (in milli seconds) for container GPU metrics")
 	gpuConfigFile                  = flag.String("gpu-config", "/etc/nvidia/gpu_config.json", "File with GPU configurations for device plugin")
+	publishDriverVersion           = flag.Bool("publish-driver-version", false, "If true, the device plugin will publish NVIDIA driver versions to the Kubernetes Node annotation")
 )
 
 func parseGPUConfig(gpuConfigFile string) (gpumanager.GPUConfig, error) {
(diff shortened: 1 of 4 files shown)


cmd/nvidia_gpu/nvidia_gpu.go
```diff
@@ -153,5 +156,27 @@ func main() {
 		defer hc.Stop()
 	}
 
+	if *publishDriverVersion {
+		glog.Infoln("Publishing NVIDIA driver version annotations to Node")
+		kubeClient, err := util.BuildKubeClient()
+		if err != nil {
+			glog.Warningf("Failed to build kube client for annotating: %v", err)
```
rawsocket: return from here, reduce nesting.

cmd/nvidia_gpu/nvidia_gpu.go
```diff
@@ -153,5 +156,27 @@ func main() {
...
+		} else {
+			nodeName := os.Getenv("NODE_NAME")
+			if nodeName == "" {
+				glog.Warning("NODE_NAME environment variable not set, skipping publishing driver version annotations")
```
rawsocket, line 168: ditto

cmd/nvidia_gpu/nvidia_gpu.go
```diff
@@ -153,5 +156,27 @@ func main() {
...
+			} else {
+				go func() {
```
rawsocket, line 173: There is a timing concern here: at this point, if main() exists quickly, go routine will be killed instantly and the annotation will never reach the node. You could use sync.WaitGroup to synchronize.

pkg/gpu/nvidia/version_visibility/version_visibility.go
```diff
@@ -0,0 +1,86 @@
...
+		annotations[DriverVersionMinor] = parts[1]
+	}
+	if len(parts) >= 3 {
+		annotations[DriverVersionRevision] = parts[2]
+	}
+
+	return annotations
+}
+
+func patchNodeAnnotations(kubeClient kubernetes.Interface, nodeName string, annotations map[string]string) error {
```
rawsocket: Pass ctx context.Context from main and use it later instead of context.Background() when calling Apply().

Create context in main to pass it down. 
    ctx, cancel := context.WithTimeout(context.Background(), apiTimeout)
    defer cancel()

    the timeout could be some arbitrary number of seconds for this operation to complete, give main() some extra time before it declares a failure.

pkg/gpu/nvidia/version_visibility/version_visibility.go
```diff
@@ -0,0 +1,86 @@
...
+	if nodeName == "" {
+		return fmt.Errorf("node name is empty")
+	}
+
+	nodeApplyConfiguration := corev1apply.Node(nodeName).
+		WithAnnotations(annotations)
+
+	glog.Infof("Applying node %s annotations using SSA: %v", nodeName, annotations)
+	_, err := kubeClient.CoreV1().Nodes().Apply(
+		context.Background(),
```
rawsocket: context.Background() is an empty context that never times out and cannot be cancelled. You need to set up a context in main and pass it down including the timeout.

rawsocket (changes requested):

cmd/nvidia_gpu/nvidia_gpu.go
leiyiz, line 164: we don't want to block the later ngm.Serve if kubeClient building fails.

after the glog.Warningf this code exists the if condition and goes to L181.

I can add a comment for readability. WDYT

cmd/nvidia_gpu/nvidia_gpu.go
leiyiz: I think we are ok here, ngm.Serve is a long running func.  
If main somehow exited the GPU device plugin container will retry.

The main concern is still that we don't want to block normal operation for main func ngm.Serve to proceed or exit

cmd/nvidia_gpu/nvidia_gpu.go
rawsocket: This was mostly about reducing the nesting in if-then-else, by early return and avoiding else {} nesting completely.

PR #605, opened 2026-05-04
Add support for automatic vGPU MIG configuration in device plugin

ankitjoju-google:
This also completely removes the nvidia-grid daemon logic.


cmd/nvidia_gpu/nvidia_gpu.go
```diff
@@ -68,13 +71,47 @@ func parseGPUConfig(gpuConfigFile string) (gpumanager.GPUConfig, error) {
...
 	if err != nil {
 		return gpumanager.GPUConfig{}, err
 	}
 	return gpuConfig, nil
 }
 
+// parseGPUFractionDivisor reads the GPU fraction divisor from the given file and returns it.
+// All GPUs have a default fraction of 1, any integer greater implies its a vGPU machine type.
+func parseGPUFractionDivisor(gpuFractionDivisorFile string) (int, error) {
+	fractionDivisor := 1
```
carreter: nit: Early return `1` if the divisor file path is an empty string for clearer happy path

cmd/nvidia_gpu/nvidia_gpu.go
```diff
@@ -68,13 +71,47 @@ func parseGPUConfig(gpuConfigFile string) (gpumanager.GPUConfig, error) {
...
+	if gpuFractionDivisorFile != "" {
+		glog.Infof("Reading GPU fraction divisor file: %s", gpuFractionDivisorFile)
+		file, err := os.ReadFile(gpuFractionDivisorFile)
+		if err != nil {
+			if os.IsNotExist(err) {
```
carreter: nit: Apparently this is no longer idiomatic, huh! [Go docs](https://pkg.go.dev/os#IsNotExist:~:text=This%20function%20predates%20errors.Is.%20It%20only%20supports%20errors%20returned%20by%20the%20os%20package.%20New%20code%20should%20use%20errors.Is(err%2C%20fs.ErrNotExist).) say to prefer `errors.Is(err, fs.ErrNotExist)`

regardless, you could restructure this to be `if os.IsNotExist(err) / else if err != nil ` and the rest outside the else to reduce nesting

cmd/nvidia_gpu/nvidia_gpu.go
```diff
@@ -68,13 +71,47 @@ func parseGPUConfig(gpuConfigFile string) (gpumanager.GPUConfig, error) {
...
+				return 1, fmt.Errorf("GPU fraction divisor file not found at %v, defaulting to 1", gpuFractionDivisorFile)
+			} else {
+				return 1, fmt.Errorf("Failed to read GPU fraction divisor file at %v, defaulting to 1: %v", gpuFractionDivisorFile, err)
+			}
+		} else {
+			gpuFractionDivisor := strings.ToLower(strings.Trim(string(file), " \r\n\x00"))
```
carreter:
```suggestion
			gpuFractionDivisor := strings.TrimSpace(string(file)
```

nit: this hits all unicode whitespace characters and is more legible. lowercase is not necessary because we fail if it's not a number

cmd/nvidia_gpu/nvidia_gpu.go
```diff
@@ -68,13 +71,47 @@ func parseGPUConfig(gpuConfigFile string) (gpumanager.GPUConfig, error) {
...
+			parsedValue, err := strconv.Atoi(gpuFractionDivisor)
+			if err != nil {
+				return 1, fmt.Errorf("Failed to parse GPU fraction divisor file at %v, defaulting to 1: %v", gpuFractionDivisorFile, err)
+			} else {
```
carreter: nit: Unnecessary else

pkg/gpu/nvidia/mig/mig.go
```diff
@@ -130,17 +130,43 @@ func (d *DeviceManager) Start(partitionSize string) error {
...
 	d.gpuPartitionSpecs = make(map[string][]pluginapi.DeviceSpec)
 
+	numPartitionsPerGPU, err := d.DiscoverMIGDevices()
+	if err != nil {
+		return err
+	}
+
+	numPartitionedGPUs := 0
+	for gpuID, numPartitions := range numPartitionsPerGPU {
+		if numPartitions != maxPartitionCount {
```
carreter, line 140: Why is this `!=` instead of `<=` here if it's a max?

carreter (commented):
Only had time to read through a bit, mostly nits here

pkg/gpu/nvidia/mig/mig.go
ankitjoju-google: This is just how it was designed beforehand, I didn't want to change any prior functionality in case something relied upon it.

cmd/nvidia_gpu/nvidia_gpu.go
```diff
@@ -75,6 +80,34 @@ func parseGPUConfig(gpuConfigFile string) (gpumanager.GPUConfig, error) {
...
+	fractionDivisor := 1
+
+	if gpuFractionDivisorFile == "" {
+		return 1, fmt.Errorf("GPU fraction divisor file not defined at %v, defaulting to 1", gpuFractionDivisorFile)
+	}
+
+	glog.Infof("Reading GPU fraction divisor file: %s", gpuFractionDivisorFile)
+	file, err := os.ReadFile(gpuFractionDivisorFile)
+	if errors.Is(err, fs.ErrNotExist) {
+		return 1, fmt.Errorf("GPU fraction divisor file not found at %v, defaulting to 1", gpuFractionDivisorFile)
```
Jiaqicao257, line 92: Put this as a log instead? This file won't exist for non vGPU types, which is majority of the fleet.

pkg/gpu/nvidia/mig/mig.go
Jiaqicao257, line 140: How's maxPartitionCount determined for vGPU?

ankitjoju-google: It's not, `DiscoverMigDevices()` only mounts the MIG devices if they're available, we don't call the actual MigManager start function for the reason that we don't actually have access to any of the MIG information on our end due to GCE obscuring it. The only way to retreive it would be to execute `nvidia-smi -L` or something and parse the output which doesn't seem like a great idea.
```
if ngm.gpuConfig.GPUPartitionSize != "" {
		if err := ngm.migDeviceManager.Start(ngm.gpuConfig.GPUPartitionSize); err != nil {
			return fmt.Errorf("failed to start mig device manager: %v", err)
		}
	} else if ngm.gpuConfig.GPUFractionDivisor > 1 {
		// This machine type is a vGPU instance, therefore the MIG devices still need to be mounted but the actual MIG device manager doesn't have to be.
		// This is because in the background, GCE provisions vGPU shapes using MIG, and so they are managing the MIG state but device plugin needs to be aware that MIGs are being used
		// otherwise CUDA and other NVIDIA utilities will be unavailable to K8S.
		glog.Info("vGPU machine type detected. Mounting MIG devices manually.")
		if _, err := ngm.migDeviceManager.DiscoverMIGDevices(); err != nil {
			return fmt.Errorf("failed to discover MIG devices: %v", err)
		}
	}
```

cmd/nvidia_gpu/nvidia_gpu.go
ankitjoju-google: What log do you mean? Is there a different logging object I should use instead?

ankitjoju-google: I moved it after the initial file checks so the log message only appears if the file exists on the machine.

cmd/nvidia_gpu/nvidia_gpu.go
```diff
@@ -75,6 +80,33 @@ func parseGPUConfig(gpuConfigFile string) (gpumanager.GPUConfig, error) {
...
+	}
+
+	gpuFractionDivisorStr := strings.TrimSpace(string(file))
+	glog.Infof("GPU fraction divisor file: %s, fraction = %s", gpuFractionDivisorFile, gpuFractionDivisorStr)
+	parsedValue, err := strconv.Atoi(gpuFractionDivisorStr)
+	if err != nil {
+		return 1, fmt.Errorf("Failed to parse GPU fraction divisor file at %v, defaulting to 1: %v", gpuFractionDivisorFile, err)
+	}
+
+	fractionDivisor = parsedValue
```
leiyiz: nit: check for `parsedValue >= 1` here

cmd/nvidia_gpu/nvidia_gpu.go
```diff
@@ -75,6 +80,33 @@ func parseGPUConfig(gpuConfigFile string) (gpumanager.GPUConfig, error) {
 	return gpuConfig, nil
 }
 
+// parseGPUFractionDivisor reads the GPU fraction divisor from the given file and returns it.
+// All GPUs have a default fraction of 1, any integer greater implies its a vGPU machine type.
+func parseGPUFractionDivisor(gpuFractionDivisorFile string) (int, error) {
+	fractionDivisor := 1
```
leiyiz: nit: you do not need to declare `fractionDivisor` here, it's unused until L106 and it was overwritten

nvidia-persistenced-installer/nvidia_persistenced_installer.go
```diff
@@ -84,14 +83,6 @@ func main() {
 		// Don't exit as this is intended for a side car which would cause it to restart infinitely.
 	}
 
-	// Read machine type from the file generated by the initContainer
```
leiyiz, line 87: do we have a public deployment yaml where we have the initContainer spec or is it internal only

pkg/gpu/nvidia/manager.go
```diff
@@ -393,6 +395,14 @@ func (ngm *nvidiaGPUManager) Start() error {
 		if err := ngm.migDeviceManager.Start(ngm.gpuConfig.GPUPartitionSize); err != nil {
 			return fmt.Errorf("failed to start mig device manager: %v", err)
 		}
+	} else if ngm.gpuConfig.GPUFractionDivisor > 1 {
+		// This machine type is a vGPU instance, therefore the MIG devices still need to be mounted but the actual MIG device manager doesn't have to be.
+		// This is because in the background, GCE provisions vGPU shapes using MIG, and so they are managing the MIG state but device plugin needs to be aware that MIGs are being used
+		// otherwise CUDA and other NVIDIA utilities will be unavailable to K8S.
+		glog.Info("vGPU machine type detected. Mounting MIG devices manually.")
+		if _, err := ngm.migDeviceManager.DiscoverMIGDevices(); err != nil {
```
leiyiz: because you are extracting and using `DiscoverMIGDevices` purely for it's side effect, consider renaming it to include hint of the side effect.

`DiscoverMIGAndMount` or something appropriate

leiyiz (commented):
lgtm in general to me.

nvidia-persistenced-installer/nvidia_persistenced_installer.go
ankitjoju-google: There is a public DS that uses it, but its only vGPU and that will be deprecated next week.

pkg/gpu/nvidia/manager.go
ankitjoju-google: Changed

leiyiz (approved):

PR #617, opened 2026-06-17
feat: add GPU UUID and Device ID to XID error events

YtG12:
Unit tests passed.
Manually validated on a live GKE cluster using the illegal-memory-access demo pod.
Screenshot 2026-06-17 at 11 56 26 AM

### DIFF
--- a/pkg/gpu/nvidia/health_check/health_checker.go
+++ b/pkg/gpu/nvidia/health_check/health_checker.go
@@ -389,12 +389,13 @@ func (hc *GPUHealthChecker) recordXIDEvent(e nvml.Event, cd callDevice) error {
 		return err
 	}
 
-	var msg string
-	if e.UUID == nil || len(*e.UUID) == 0 {
-		msg = fmt.Sprintf("Caught XID error, XID=%d", e.Edata)
-	} else {
+	var sb strings.Builder
+	fmt.Fprintf(&sb, "Caught XID error, XID=%d", e.Edata)
+
+	if e.UUID != nil && len(*e.UUID) > 0 {
 		var affectedGpuIDs []string
 		var affectedGpuUUIDs []string
+
 		for _, d := range hc.devices {
 			nvmlDev, ok := hc.nvmlDevices[d.ID]
 			if !ok || nvmlDev == nil {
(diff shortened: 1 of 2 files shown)


pkg/gpu/nvidia/health_check/health_checker.go
```diff
@@ -383,12 +383,45 @@ func (hc *GPUHealthChecker) updateLastHeartbeatTime() {
...
+			}
+		}
+		if len(affectedGpuIDs) > 0 {
+			msg = fmt.Sprintf("Caught XID error, XID=%d, GPU UUID=%s, Device ID=%s", e.Edata, strings.Join(affectedGpuUUIDs, ", "), strings.Join(affectedGpuIDs, ", "))
+		} else {
+			msg = fmt.Sprintf("Caught XID error, XID=%d, GPU UUID=%s", e.Edata, *e.UUID)
+		}
+	}
+
+	hc.recorder.Eventf(node, v1.EventTypeWarning, "XIDError", msg)
```
ankitjoju-google, line 428: nit: You might want to use string builder instead and make it so that the string is built up progressively with the data that's available rather than redeclaring the `msg` for each branch.

Ex:
```go
var sb strings.Builder

// 1. Write the base message that applies to all conditions
fmt.Fprintf(&sb, "Caught XID error, XID=%d", e.Edata)

if e.UUID != nil && len(*e.UUID) > 0 {
	var affectedGpuIDs []string
	var affectedGpuUUIDs []string
    
	for _, d := range hc.devices {
		nvmlDev, ok := hc.nvmlDevices[d.ID]
		if !ok || nvmlDev == nil {
			continue
		}
		uuid := nvmlDev.UUID
		gpu, gi, ci, err := cd.parseMigDeviceUUID(uuid)
		if err != nil {
			gpu = uuid
			gi = 0xFFFFFFFF
			ci = 0xFFFFFFFF
		}

		if e.GpuInstanceId != nil && e.ComputeInstanceId != nil &&
			gpu == *e.UUID && gi == *e.GpuInstanceId && ci == *e.ComputeInstanceId {
			affectedGpuIDs = append(affectedGpuIDs, d.ID)
			affectedGpuUUIDs = append(affectedGpuUUIDs, uuid)
		}
	}

	// 2. Progressively tack on the extra details based on what was found
	if len(affectedGpuIDs) > 0 {
		fmt.Fprintf(&sb, ", GPU UUID=%s", strings.Join(affectedGpuUUIDs, ", "))
		fmt.Fprintf(&sb, ", Device ID=%s", strings.Join(affectedGpuIDs, ", "))
	} else {
		fmt.Fprintf(&sb, ", GPU UUID=%s", *e.UUID)
	}
}

// 3. Extract the final compiled string
msg := sb.String()
```

YtG12: Thanks for the suggestion! updated.

ankitjoju-google (approved):

PR #632, opened 2026-08-24
Swap gpu-fraction-divisor path to use underscores instead of hyphens

ankitjoju-google:
b/544643611


leiyiz (approved):
can you link the bug in the PR description for posterity

